PR #2, opened 2017-03-11
add initial version of simple property service

brharrington:
This is a simple property service backed by a DynamoDB table
that can be used with `iep-module-archaius2`. This service
currently only supports the read operations. Writes are
managed by a deployment service internally.


brharrington:
Note: build failure was due to branch being local and failing to publish, not the change itself.

PR #42, opened 2018-01-26
fix key used for fallback expr summary

brharrington:
This should be the group name not the multiple
possible matches selector.

### DIFF
--- a/iep-lwc-bridge/src/main/scala/com/netflix/iep/lwc/ExprUpdateActor.scala
+++ b/iep-lwc-bridge/src/main/scala/com/netflix/iep/lwc/ExprUpdateActor.scala
@@ -88,7 +88,7 @@ class ExprUpdateActor(config: Config, registry: Registry, evaluator: Evaluator)
     * will be evaluated for every datapoint.
     */
   private def updateEvaluator(exprs: SubscriptionList): Unit = {
-    val appGroups = new java.util.HashMap[String, SubscriptionList]()
+    val groups = new java.util.HashMap[String, SubscriptionList]()
     val iter = exprs.iterator()
     while (iter.hasNext) {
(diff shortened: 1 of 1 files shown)


iep-lwc-bridge/src/main/scala/com/netflix/iep/lwc/ExprUpdateActor.scala
```diff
@@ -88,24 +88,24 @@ class ExprUpdateActor(config: Config, registry: Registry, evaluator: Evaluator)
...
       val name = exactTags.getOrDefault("name", ManyPossibleMatches)
       val group = s"$app:$name"
-      val subs = appGroups.computeIfAbsent(group, _ => new java.util.ArrayList[Subscription]())
+      val subs = groups.computeIfAbsent(group, _ => new java.util.ArrayList[Subscription]())
       subs.add(sub)
     }
-    appGroups.forEach { (app, subs) =>
-      if (app == ManyPossibleMatches)
+    groups.forEach { (group, subs) =>
+      if (group == AllGroup)
```
zimmermatt: Is there enough of a semantic difference that the scaladoc update made in pr #41 be updated as well?

brharrington: Possibly, I as in a hurry to see if it would unblock something, I probably need to do a cleanup pass. That said, it didn't work as well as I was hoping so I'll probably go ahead and refactor this to use the QueryIndex from Atlas. Originally I was trying to avoid this to keep it as similar to what would run on the client as possible and I think the QueryIndex is too costly there. Now though I think there are some easier optimizations we can use in the simple client case that do not work for the bridge.

PR #46, opened 2018-02-13
pass query in as filter to druid group by

brharrington:
When fetching the data form Druid the query was being
used to select the datasource and name, but was ignored
after that. Now it is converted to a filter and passed
to the group by request that is sent to Druid.


brharrington:
Build failure is due to unrelated issue, see #47.

PR #81, opened 2018-10-07
Initial version of SES monitoring service

zimmermatt:
This service consumes SES sending activity notifications and publishes
key aspects as metrics. It additionally logs the full notification to
disk.

### DIFF
--- a/iep-ses-monitor/src/main/scala/com/netflix/iep/ses/AppModule.scala
+++ b/iep-ses-monitor/src/main/scala/com/netflix/iep/ses/AppModule.scala
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2014-2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.iep.ses
+
+import com.amazonaws.PredefinedClientConfigurations
+import com.amazonaws.auth.DefaultAWSCredentialsProviderChain
+import com.amazonaws.services.sqs.AmazonSQSAsync
+import com.amazonaws.services.sqs.AmazonSQSAsyncClient
+import com.google.inject.AbstractModule
+import com.google.inject.Provides
+import com.google.inject.multibindings.Multibinder
+import com.netflix.iep.NetflixEnvironment
+import com.netflix.iep.service.Service
+
+class AppModule extends AbstractModule {
+  override def configure(): Unit = {
+    val serviceBinder = Multibinder.newSetBinder(binder(), classOf[Service])
+    serviceBinder.addBinding().to(classOf[SesMonitoringService])
+  }
+
+  @Provides
+  def provideAwsSqsAsyncClient(): AmazonSQSAsync = {
+    // TODO add to AwsClientFactory?
+    AmazonSQSAsyncClient
+      .asyncBuilder()
+      .withRegion(NetflixEnvironment.region())
+      .withCredentials(new DefaultAWSCredentialsProviderChain)
+      .withClientConfiguration(PredefinedClientConfigurations.defaultConfig().withGzip(true))
+      .build()
+  }
+}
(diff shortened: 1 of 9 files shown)


iep-ses-monitor/src/main/scala/com/netflix/iep/ses/SesMonitoringService.scala
```diff
@@ -0,0 +1,192 @@
...
+  ) = {
+
+    // getOrElse defensively since notifications are an external input we don't control
+    val complaintDetails =
+      notification.getOrElse("complaint", Map.empty[String, Any]).asInstanceOf[Map[String, Any]]
+
+    // complaintType is specified as optional
+    val complaintTypeKey = "complaintFeedbackType"
+    val complaintTypeValue =
+      complaintDetails.getOrElse(complaintTypeKey, "unknown").asInstanceOf[String]
```
brharrington: Consider moving the path lookup to a helper function rather than repeating each time. It would make it easer to follow the important logic of the function. Something like:

```scala
  @scala.annotation.tailrec
  private def getPath(obj: Map[String, Any], path: String*): String = {
    path.toList match {
      case ks if ks.isEmpty || obj.isEmpty =>
        "unknown"
      case k :: Nil =>
        obj.getOrElse(k, "unknown").asInstanceOf[String]
      case k :: ks =>
        val subObj = obj.getOrElse(k, Map.empty[String, Any]).asInstanceOf[Map[String, Any]]
        getPath(subObj, ks: _*)
    }
  }

  val complaintTypeValue = getPath(notification, "complaint", "complaintFeedbackType")
```

iep-ses-monitor/src/main/scala/com/netflix/iep/ses/SesMonitoringService.scala
```diff
@@ -0,0 +1,192 @@
...
+  ) = {
+
+    // getOrElse defensively since notifications are an external input we don't control
+    val bounceDetails =
+      notification.getOrElse("bounce", Map.empty[String, Any]).asInstanceOf[Map[String, Any]]
+
+    val bounceType = bounceDetails.getOrElse("bounceType", "unknown").asInstanceOf[String]
+    val bounceSubType = bounceDetails.getOrElse("bounceSubType", "unknown").asInstanceOf[String]
+
+    Vector(new BasicTag("type", bounceType), new BasicTag("subType", bounceSubType))
```
brharrington: The key `type` would get overridden with servo. I don't think it will cause a problem here because spectator is no longer routed through servo internally, but historically it was advisable to avoid this key.

zimmermatt: It's not clear to me whether or not you're requesting a change. I wanted to keep this dimension consistent across `notificationType`'s to ease join/group by.

iep-ses-monitor/src/main/scala/com/netflix/iep/ses/AppModule.scala
```diff
@@ -0,0 +1,44 @@
...
+
+class AppModule extends AbstractModule {
+  override def configure(): Unit = {
+    val serviceBinder = Multibinder.newSetBinder(binder(), classOf[Service])
+    serviceBinder.addBinding().to(classOf[SesMonitoringService])
+  }
+
+  @Provides
+  def provideAwsSqsAsyncClient(): AmazonSQSAsync = {
+    // TODO add to AwsClientFactory?
```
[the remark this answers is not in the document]
zimmermatt: Yeah. It's unfortunate they block behind the async interface. The alpakka SQS connector requires an implicit `AmazonSQSAsync` instance in scope. It deals with the blocking nature of `AmazonSQSAsync` and correctly handles the callbacks with `akka.stream.stage.GraphStageLogic#getAsyncCallback` in its `GraphStageLogic` implementation. See:
* https://developer.lightbend.com/docs/alpakka/1.0-M1/sqs.html#setup
* https://github.com/akka/alpakka/blob/v1.0-M1/sqs/src/main/scala/akka/stream/alpakka/sqs/impl/SqsSourceStage.scala

I figured this would be a good opportunity to experiment with alpakka and the AWS v1 SDK since SQS in the v2 SDK is still considered preview. I didn't see any anticipated timeline for a production ready v2 SQS.

If the alpakka implementation with the v1 `AmazonSQSAsync` doesn't work well in practice, I'll either implement an akka stream source with the v1 `AmazonSQS` (non-async) and/or test implementing one with the v2 async.

zimmermatt:
btw... I'm not sure what's different about the config of this project (or maybe it at the Netflix-Skunkworks level), but the function of requesting a reviewer doesn't seem to be available.

brharrington:
Works for me, I'm guessing you are not a member of the org.

iep-ses-monitor/src/main/scala/com/netflix/iep/ses/SesMonitoringService.scala
brharrington: Keeping it consistent is fine, but you may run into trouble using the key `type`.

iep-ses-monitor/src/main/scala/com/netflix/iep/ses/AppModule.scala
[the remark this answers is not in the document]
brharrington: Alpakka doesn't have a stable release yet either. Since this is a newer app it seems like a reasonable place to experiment with the v2 SDK.

PR #95, opened 2018-10-26
Report `bouncedRecipients` list

zimmermatt:
One of our users needs the bounced recipients for their use case. This
commit adds extraction of the `bouncedRecipients` list and reports the
20 most frequent addresses.

### DIFF
--- a/iep-ses-monitor/src/main/scala/com/netflix/iep/ses/SesMonitoringService.scala
+++ b/iep-ses-monitor/src/main/scala/com/netflix/iep/ses/SesMonitoringService.scala
@@ -58,6 +58,7 @@ class SesMonitoringService @Inject()(
 
   private val notificationsId = registry.createId("ses.monitor.notifications")
   private val sourceEmailLimiter = CardinalityLimiters.mostFrequent(20)
+  private val bouncedRecipientLimiter = CardinalityLimiters.mostFrequent(20)
 
   private val streamFailures = registry.counter("ses.monitor.streamFailures")
 
@@ -113,8 +114,7 @@ class SesMonitoringService @Inject()(
       val json = Json.decode[Map[String, Any]](message.getBody)
       val messageObject =
         Json.decode[Map[String, Any]](json.getOrElse("Message", "{}").asInstanceOf[String])
-      val tags = extractTags(messageObject)
-      registry.counter(notificationsId.withTags(tags: _*)).increment()
+      incrementCounter(messageObject)
     } catch {
       case e: Exception =>
         registry.counter(deserializationFailuresId.withTag("exception", e.getClass.getSimpleName))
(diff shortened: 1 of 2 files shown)


iep-ses-monitor/src/main/scala/com/netflix/iep/ses/SesMonitoringService.scala
```diff
@@ -127,50 +127,74 @@ class SesMonitoringService @Inject()(
     MessageAction.delete(message)
   }
 
+  private def getPathAsString(obj: Map[String, Any], path: String*) = {
```
brharrington:
```suggestion
  private def getPathAsString(obj: Map[String, Any], path: String*): String = {
```

It is generally better to be explicit with return types on the method. Otherwise it can change in strange ways due to inference. Being explicit provides a sanity check from the compiler.

iep-ses-monitor/src/main/scala/com/netflix/iep/ses/SesMonitoringService.scala
```diff
@@ -127,50 +127,74 @@ class SesMonitoringService @Inject()(
...
+    getPath(obj, path: _*).fold("unknown")(_.asInstanceOf[String])
+  }
+
+  private def getPathAsList(obj: Map[String, Any], listElementKey: String, pathToList: String*) = {
```
brharrington:
```suggestion
  private def getPathAsList(obj: Map[String, Any], listElementKey: String, pathToList: String*): List[String] = {
```

iep-ses-monitor/src/main/scala/com/netflix/iep/ses/SesMonitoringService.scala
```diff
@@ -127,50 +127,74 @@ class SesMonitoringService @Inject()(
...
+        obj.get(k)
       case k :: ks =>
         val subObj = obj.getOrElse(k, Map.empty[String, Any]).asInstanceOf[Map[String, Any]]
         getPath(subObj, ks: _*)
     }
   }
 
-  private[ses] def extractTags(notification: Map[String, Any]) = {
-
+  private[ses] def incrementCounter(notification: Map[String, Any]) = {
```
brharrington:
```suggestion
  private[ses] def incrementCounter(notification: Map[String, Any]): Unit = {
```

zimmermatt:
Not sure if it's flaky or not, but a test failed in an unmodified sub-project (`iep-lwc-cloudwatch`) failed in the `openjdk11` build. Rerunning it to see.

```
[info] *** 1 TEST FAILED ***
[error] Failed tests:
[error] 	com.netflix.iep.lwc.ForwardingServiceSuite
```

brharrington:
Which test failed and what was the specific error?

iep-ses-monitor/src/main/scala/com/netflix/iep/ses/SesMonitoringService.scala
zimmermatt: Happy to add it. Most style guides I've worked with favor inference on `private` methods and explicit on public, unless it's ambiguous by nature.

I see now that our projects do consistently specify a type, regardless of visibility.

zimmermatt:
> Which test failed and what was the specific error?

Sorry, didn't capture that as I didn't realize that PR builds don't maintain history.

Pretty sure it was `test("toCloudWatchPut: multiple accounts")`

PR #98, opened 2018-11-03
Json schema validation for cluster config

narayaruna:


project/Dependencies.scala
```diff
@@ -62,6 +62,7 @@ object Dependencies {
   val jacksonSmile2      = "com.fasterxml.jackson.dataformat" % "jackson-dataformat-smile" % jackson
   val jodaConvert        = "org.joda" % "joda-convert" % "1.8.1"
   val jol                = "org.openjdk.jol" % "jol-core" % "0.5"
+  val jsonSchema         = "com.github.everit-org.json-schema" % "org.everit.json.schema" % "1.9.2"
```
brharrington: Consider using the [json-schema-validator](https://github.com/java-json-tools/json-schema-validator) library instead. That is what configbin uses so it will help make sure the interpretation of the schema will match.

iep-lwc-cloudwatch/src/main/resources/cluster-config-schema.json
```diff
@@ -0,0 +1,67 @@
...
+        "properties": {
+          "metricName": {
+            "description": "Name of the metric that will be created in CloudWatch",
+            "type": "string",
+            "pattern": "^[a-zA-Z0-9]+[a-zA-Z0-9_\\-\\.]*[a-zA-Z0-9]+$"
+          },
+          "atlasUri": {
+            "description": "Atlas query",
+            "type": "string",
+            "format": "uri"
```
brharrington: How strict is this? We tend to allow lenient parsing of URIs for convenience. For example, you can use a regex with an or, `|`, that would fail with `URI.parse` because it is not escaped.

iep-lwc-cloudwatch/src/main/resources/cluster-config-schema.json
```diff
@@ -0,0 +1,67 @@
...
+              "properties": {
+                "name": {
+                  "description": "Name of the CloudWatch dimension",
+                  "type": "string",
+                  "pattern": "^[a-zA-Z0-9]+[a-zA-Z0-9_\\-\\.]*[a-zA-Z0-9]+$"
+                },
+                "value": {
+                  "description": "Atlas tag. Example: $(nf.asg)",
+                  "type": "string",
+                  "pattern": "^\\$\\([a-zA-Z0-9]+[a-zA-Z0-9\\.]*[a-zA-Z0-9]+\\)$"
```
brharrington: Though it is often a substitution, it doesn't have to  be. Some for example will hardcode a value so they can scale a different ASG than is publishing the metric.

iep-lwc-cloudwatch/src/main/resources/cluster-config-schema.json
```diff
@@ -0,0 +1,67 @@
...
+                }
+              }
+            }
+          },
+          "account": {
+            "description": "AWS account to send the metric to. Typically $(nf.account)",
+            "type": "string",
+            "pattern": "^\\$\\([a-zA-Z0-9]+[a-zA-Z0-9\\.]*[a-zA-Z0-9]+\\)$"
```
brharrington: Similar to comment on the value, they could also use the explicit account id string.

iep-lwc-cloudwatch/src/test/scala/com/netflix/iep/lwc/JsonSchemaSuite.scala
```diff
@@ -0,0 +1,214 @@
...
+      "#/expressions/0/dimensions/0/value: string [asg] does not match pattern ^\\$\\([a-zA-Z0-9]+[a-zA-Z0-9\\.]*[a-zA-Z0-9]+\\)$"
+    )
+  }
+
+  test("Fail for invalid account variable") {
+    val config = makeConfig(account = "account")
+    assertException(
+      schema.validate(new JSONObject(config)),
+      "#/expressions/0/account: string [account] does not match pattern ^\\$\\([a-zA-Z0-9]+[a-zA-Z0-9\\.]*[a-zA-Z0-9]+\\)$"
+    )
```
brharrington: There is a built-in intercept function you could use:

```scala
val e = intercept[ValidationException] {
  schema.validate(new JSONObject(config))
}
assert(e.getMessage === "...")
```

iep-lwc-cloudwatch/src/main/resources/cluster-config-schema.json
```diff
@@ -0,0 +1,67 @@
...
+      "format": "email"
+    },
+    "expressions": {
+      "description": "List of metric expressions",
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "items": {
+        "type": "object",
+        "required": [ "metricName", "atlasUri", "dimensions", "account" ],
```
brharrington: `dimensions` is optional

iep-lwc-cloudwatch/src/main/resources/cluster-config-schema.json
```diff
@@ -0,0 +1,67 @@
...
+            "type": "string",
+            "format": "uri"
+          },
+          "comment": {
+            "type": "string"
+          },
+          "dimensions": {
+            "description": "Details about AWS CloudWatch dimension that will be created",
+            "type": "array",
+            "minItems": 1,
```
brharrington: Since the default is an empty list with no dimensions, we should allow them to use an explicit empty list.

iep-lwc-cloudwatch/src/test/scala/com/netflix/iep/lwc/JsonSchemaSuite.scala
```diff
@@ -0,0 +1,214 @@
...
+      schema.validate(new JSONObject(config)),
+      "#/expressions/0/metricName: string [.nodejs.cpuUsage] does not match pattern ^[a-zA-Z0-9]+[a-zA-Z0-9_\\-\\.]*[a-zA-Z0-9]+$"
+    )
+  }
+
+  test("Fail for invalid atlasUri name") {
+    val config = makeConfig(atlasUri = "http://localhost/api/v1/graph?q=^query")
+    assertException(
+      schema.validate(new JSONObject(config)),
+      "#/expressions/0/atlasUri: [http://localhost/api/v1/graph?q=^query] is not a valid URI"
```
brharrington: Re comment above about URI parsing, we tend to be more lenient. Strict parsing means the user would likely get some errors if they just copy and paste a URI from some of the tools.

```scala
println(Uri("/api/v1/graph?q=^query", Uri.ParsingMode.Relaxed).query())
```

Yields:

```
q=%5Equery
```

brharrington:
Note, you should be able to run `make format` to fix the scalafmt errors causing the build to fail.

PR #101, opened 2018-12-06
A standalone app to validate and fix Cloudwatch forwarding configs

narayaruna:


project/Dependencies.scala
```diff
@@ -62,7 +62,7 @@ object Dependencies {
   val jacksonSmile2      = "com.fasterxml.jackson.dataformat" % "jackson-dataformat-smile" % jackson
   val jodaConvert        = "org.joda" % "joda-convert" % "1.8.1"
   val jol                = "org.openjdk.jol" % "jol-core" % "0.5"
-  val jsonSchema         = "com.github.java-json-tools" % "json-schema-validator" % "2.2.10"
+  val jsonSchema         = "com.github.java-json-tools" % "json-schema-validator" % "2.2.8"
```
brharrington: Why move back to an older version?

iep-lwc-cloudwatch/src/main/scala/com/netflix/iep/lwc/admin/ManageConfig.scala
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  validateAllConfigs()
+
+//  fixConfigs(
+//    _.exists(
+//      _ == "object has missing required properties ([\"email\"])"
+//    ),
+//    setEmail(_)
+//  )
```
brharrington: It seems like this is more of a script and you would need to change the code to use it. Also it seems like it isn't really specific to the CloudWatch forwarding, but any ConfigBin usage that happens to have a schema.

For validating all configs against the schema, I wonder if we could get them to add that as an API to ConfigBin so it is easy to do for anything.

For the rewriting aspect, I'm not sure if this is the right place. It isn't really part of the app and would require a custom function to use. So it is more of a convenient script we just need to put somewhere.

project/Dependencies.scala
narayaruna: Pinning to the same version that Configbin uses. I did this while troubleshooting the schema validation errors. It should work even with the latest version.

iep-lwc-cloudwatch/src/main/scala/com/netflix/iep/lwc/admin/ManageConfig.scala
narayaruna: True. I created this as a script initially under `src/main/scripts`. But couldn't get the script pickup the dependencies correctly. 
I intent to keep this someplace so it is version control and enhance it as we write more utilities to manage configurations. I see following alternate locations for this:
- Sort out sbt issues and keep the script under `src/main/scripts`
- Keep it in this project but move it under `src/test/scala`
- Move this to `insight-pkg` under `iep-lwc-cloudwatch/src/test`
- Create a new subproject `iep-configbin-utils` under `iep-apps`
Does the last option sounds more reasonable? Because this can be a placeholder for any utility script we might write for managing configs in configbin.

narayaruna:
Moved this to the scripts repo in Stash.

PR #118, opened 2019-02-22
Report metric forwarding status to Admin service

narayaruna:


iep-lwc-cloudwatch/src/main/scala/com/netflix/iep/lwc/ForwardingService.scala
```diff
@@ -300,39 +305,75 @@ object ForwardingService extends StrictLogging {
...
+        }
       }
       .mergeSubstreams
   }
 
+  def sendToAdmin(
+    adminUri: URI,
+    client: Client
+  )(
+    implicit mat: ActorMaterializer,
```
brharrington: Ideally the stream blueprint would be created without using a specific materializer.

PR #128, opened 2019-03-15
Add endpoint for removing purge eligible expressions

narayaruna:
Adds an endpoint to update or remove cluster configurations
in ConfigBin for the given expressions.

Modified the cluster config schema to allow empty list
for `checksToSkip` field. This is convenient because Json
encoding creates this empty list.

Moved ConfigBinVersion to the shared library.

### DIFF
--- a/iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ExpressionDetailsDao.scala
+++ b/iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ExpressionDetailsDao.scala
@@ -15,7 +15,9 @@
  */
 package com.netflix.iep.lwc.fwd.admin
 
+import com.amazonaws.client.builder.AwsClientBuilder
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDB
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClientBuilder
 import com.amazonaws.services.dynamodbv2.document.DynamoDB
 import com.amazonaws.services.dynamodbv2.document.Item
 import com.amazonaws.services.dynamodbv2.document.PrimaryKey
@@ -34,6 +36,7 @@ trait ExpressionDetailsDao {
   def scan(): List[ExpressionId]
   def queryPurgeEligible(now: Long, events: List[String]): List[ExpressionId]
   def delete(id: ExpressionId): Unit
+  def isPurgeEligible(ed: ExpressionDetails, now: Long): Boolean
 }
 
 class ExpressionDetailsDaoImpl @Inject()(
@@ -78,7 +81,9 @@ class ExpressionDetailsDaoImpl @Inject()(
         i.getNumber(Timestamp).longValue(),
         Option(i.getString(FwdMetricInfoAttr)).map(Json.decode[FwdMetricInfo](_)),
         Option(i.getString(Error)).map(Json.decode[Throwable](_)),
-        i.getMap[Long](Events).asScala.toMap,
+        i.getMap[java.math.BigDecimal](Events).asScala.toMap.map {
+          case (k, v) => (k, BigDecimal(v).toLong)
+        },
         Option(i.getString(ScalingPolicy)).map(Json.decode[JsonNode](_))
       )
     }
@@ -133,6 +138,10 @@ class ExpressionDetailsDaoImpl @Inject()(
     table.deleteItem(new PrimaryKey(ExpressionIdAttr, Json.encode(id)))
   }
 
+  override def isPurgeEligible(ed: ExpressionDetails, now: Long): Boolean = {
+    ed.isPurgeEligible(now, ageLimitMillis)
+  }
+
 }
 
 case class ExpressionDetails(
@@ -142,7 +151,16 @@ case class ExpressionDetails(
   error: Option[Throwable],
   events: Map[String, Long],
   scalingPolicy: Option[JsonNode]
-)
+) {
+  import ExpressionDetails._
+
+  def isPurgeEligible(now: Long, ageLimitMillis: Long): Boolean = {
+    val ageThreshold = now - ageLimitMillis
+    events
+      .filter { case (k, _) => PurgeMarkerEvents.contains(k) }
+      .exists { case (_, v) => v < ageThreshold }
+  }
+}
 
 object ExpressionDetails {
   val TableName = "iep.lwc.fwd.cw.ExpressionDetails"
--- a/iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ExpressionDetailsDao.scala
+++ b/iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ExpressionDetailsDao.scala
@@ -15,9 +15,7 @@
  */
 package com.netflix.iep.lwc.fwd.admin
 
-import com.amazonaws.client.builder.AwsClientBuilder
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDB
-import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClientBuilder
 import com.amazonaws.services.dynamodbv2.document.DynamoDB
 import com.amazonaws.services.dynamodbv2.document.Item
 import com.amazonaws.services.dynamodbv2.document.PrimaryKey
--- a/iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/Purger.scala
+++ b/iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/Purger.scala
@@ -1,0 +1,180 @@
+/*
+ * Copyright 2014-2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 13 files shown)


iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/Purger.scala
```diff
@@ -0,0 +1,180 @@
...
+object PurgerImpl extends StrictLogging {
+
+  type Client = Flow[(HttpRequest, AccessLogger), (Try[HttpResponse], AccessLogger), NotUsed]
+
+  def getClusterConfig(
+    cwExprUri: String,
+    client: Client
+  ): Flow[String, ClusterCfgPayload, NotUsed] = {
+    Flow[String]
+      .map(key => HttpRequest(HttpMethods.GET, Uri(cwExprUri.format(key))))
```
brharrington: Will any of these need to be encoded before substitution?

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/Purger.scala
```diff
@@ -0,0 +1,180 @@
...
+      .map {
+        case ((k, e), c) =>
+          (
+            k,
+            removeExpressions(
+              e.map(_.expression),
+              c.payload
+            ).map(
+              ClusterCfgPayload(
+                c.version.copy(user = Some("admin"), comment = Some("Cleanup")),
```
brharrington: Can we provide a better comment that gives some context about why it was cleaned up?

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/Purger.scala
```diff
@@ -0,0 +1,180 @@
...
+  ): Flow[(String, Option[ClusterCfgPayload]), NotUsed, NotUsed] = {
+    Flow[(String, Option[ClusterCfgPayload])]
+      .map {
+        case (k, c) =>
+          c.map { cfg =>
+              HttpRequest(
+                HttpMethods.PUT,
+                Uri(cwExprUri.format(k)),
+                entity = Json.encode(cfg)
+              ).withHeaders(RawHeader("conditional", "true"))
```
brharrington: What is this header for?

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/Purger.scala
```diff
@@ -0,0 +1,180 @@
...
+            }
+            .getOrElse(
+              HttpRequest(
+                HttpMethods.DELETE,
```
brharrington: If I remember correctly, once the file has been deleted we lose the history and ability to restore in config bin. We might need to be more careful before deleting the file.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/Purger.scala
```diff
@@ -0,0 +1,180 @@
...
+                Uri(s"${cwExprUri.format(k)}?user=admin&comment=cleanup")
```
brharrington: Similar question about more context for the comment.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/Purger.scala
narayaruna: Sure. Let me test this. Nq ones will have trouble.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/Purger.scala
narayaruna: This is for compare & set. To make sure, Admin service is not overwriting any updates made to the config when the cleanup is in progress.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/Purger.scala
narayaruna: We could remove the expressions and leave the config or we could write the deleted config to the logs. I also thought about using an audit log table but deferred that approach for future improvements.

narayaruna:
I realize one more step is pending in the purge process which is to get rid of the DynamoDB entry.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ExpressionDetailsDao.scala
```diff
@@ -78,7 +81,9 @@ class ExpressionDetailsDaoImpl @Inject()(
         i.getNumber(Timestamp).longValue(),
         Option(i.getString(FwdMetricInfoAttr)).map(Json.decode[FwdMetricInfo](_)),
         Option(i.getString(Error)).map(Json.decode[Throwable](_)),
-        i.getMap[Long](Events).asScala.toMap,
+        i.getMap[java.math.BigDecimal](Events).asScala.toMap.map {
```
brharrington: Why BigDecimal?

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ExpressionDetailsDao.scala
```diff
@@ -142,7 +151,16 @@ case class ExpressionDetails(
...
   events: Map[String, Long],
   scalingPolicy: Option[JsonNode]
-)
+) {
+  import ExpressionDetails._
+
+  def isPurgeEligible(now: Long, ageLimitMillis: Long): Boolean = {
+    val ageThreshold = now - ageLimitMillis
+    events
+      .filter { case (k, _) => PurgeMarkerEvents.contains(k) }
```
brharrington: Are there other events that are not purge markers?

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ExpressionDetailsDao.scala
```diff
@@ -142,7 +151,16 @@ case class ExpressionDetails(
...
+      .exists { case (_, v) => v < ageThreshold }
```
brharrington: Is just one matching always sufficient? For example, if the no data policy is present, but scaling policy marker is missing, then should it be eligible?

iep-lwc-fwding-admin/src/test/scala/com/netflix/iep/lwc/fwd/admin/ExpressionDetailsDaoSuite.scala
```diff
@@ -143,3 +143,62 @@ class ExpressionDetailsDaoSuite extends FunSuite with BeforeAndAfter with Strict
...
+  test("Not eligible for purge for unknown events") {
+    val reportTs = 1551820461000L
+    val timestampThen = reportTs + 11.minutes.toMillis
+
+    val ed = new ExpressionDetails(
+      ExpressionId("config1", ForwardingExpression("", "", None, "", Nil)),
+      reportTs,
+      None,
+      None,
+      Map("foo" -> reportTs),
```
brharrington: Is an unknown event supposed to prevent it from being purge eligible or just get ignored? If I understand the logic if there was a single recognized event it would be eligible.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ExpressionDetailsDao.scala
narayaruna: getMap expects BigDecimal type when the Map value is a number. Here is a note from the Javadoc of this method.
>      * Note that any numeric type of a
>      * map is always canonicalized into <code>BigDecimal</code>, and therefore
>      * if <code>T</code> referred to a <code>Number</code> type, it would need
>      * to be <code>BigDecimal</code> to avoid a class cast exception.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ExpressionDetailsDao.scala
narayaruna: No. I have this to make it explicit for purge marker events.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ExpressionDetailsDao.scala
narayaruna: Yes, I think it should be. Any purge marker event should make the expression eligible for removal. May be we can tweak the behavior after experimenting with real data. These are triggered via API endpoints, so we have control over what to remove.

iep-lwc-fwding-admin/src/test/scala/com/netflix/iep/lwc/fwd/admin/ExpressionDetailsDaoSuite.scala
narayaruna: There are no marker events created other than the PurgeMarkerEvents now. This test I added to make sure we don't remove the expression if another event is added in future, for example an event like `FirstReportTs`.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/Purger.scala
narayaruna: I tested with ~ and ^ in the config name. Works fine.

PR #130, opened 2019-04-03
Scaling policy lookup for expressions

narayaruna:
Query Edda to determine whether a scaling policy is
created in Ec2 or Titus for the metric that is forwarded
to CW.

### DIFF
--- a/iep-lwc-fwding-admin/src/main/resources/application.conf
+++ b/iep-lwc-fwding-admin/src/main/resources/application.conf
@@ -5,6 +5,14 @@ atlas.akka {
     "com.netflix.atlas.akka.HealthcheckApi",
     "com.netflix.iep.lwc.fwd.admin.Api"
   ]
+
+  actors = ${?atlas.akka.actors} [
+    {
+      name = "scalingPolicies"
+      class = "com.netflix.iep.lwc.fwd.admin.ScalingPolicies"
+    }
+  ]
+
 }
 
 blocking-dispatcher {
@@ -18,5 +26,13 @@ blocking-dispatcher {
 iep.lwc.fwding-admin {
   age-limit = 10m
   queue-size-limit = 10000
+  edda-cache-refresh-interval = 30m
   cw-expr-uri = "http://localhost:7102/api/v2/cloudwatch-forwarding/clusters/%s"
+
+  // The host name in the properties 'cw-alarms-uri', 'ec2-policies-uri' and 'titus-policies-uri'
+  // uses the following pattern to lookup the appropriate Edda deployment.
+  // "prefix-%s.%s.%s.suffix".format(account, region, env)
+  cw-alarms-uri = "http://localhost:7103/api/v2/aws/alarms;namespace=NFLX/EPIC;_expand:(alarmName,metricName,dimensions:(name,value))"
+  ec2-policies-uri = "http://localhost:7103/api/v2/aws/scalingPolicies;_expand:(policyName,alarms:(alarmName))"
+  titus-policies-uri = "http://localhost:7103/api/v3/netflix/titusscalingPolicies;_expand:(jobId,id,scalingPolicy:(targetPolicyDescriptor:(customizedMetricSpecification)))"
 }
--- a/iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/MarkerService.scala
+++ b/iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/MarkerService.scala
@@ -15,12 +15,16 @@
  */
 package com.netflix.iep.lwc.fwd.admin
 
+import java.util.concurrent.TimeoutException
+
 import akka.NotUsed
+import akka.actor.ActorSelection
 import akka.actor.ActorSystem
 import akka.stream._
 import akka.stream.scaladsl.Flow
 import akka.stream.scaladsl.Keep
 import akka.stream.scaladsl.Sink
+import akka.stream.scaladsl.Source
 import com.netflix.atlas.akka.StreamOps
 import com.netflix.atlas.akka.StreamOps.SourceQueue
 import com.netflix.iep.aws.AwsClientFactory
@@ -35,6 +39,12 @@ import javax.inject.Inject
 import scala.util.Failure
 import scala.util.Success
 import scala.util.Try
+import akka.pattern.ask
+import akka.util.Timeout
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.concurrent.duration._
 
 trait MarkerService {
   var queue: SourceQueue[Report]
@@ -53,17 +63,31 @@ class MarkerServiceImpl @Inject()(
   import MarkerServiceImpl._
 
   private implicit val mat = ActorMaterializer()
+  private implicit val ec = scala.concurrent.ExecutionContext.global
   private var killSwitch: KillSwitch = _
 
   override var queue: SourceQueue[Report] = _
 
   override def startImpl(): Unit = {
     val sizeLimit = config.getInt("iep.lwc.fwding-admin.queue-size-limit")
+    val scalingPolicies = system.actorSelection("/user/scalingPolicies")
 
     val (q, k) = StreamOps
       .blockingQueue[Report](registry, "fwdingAdminCwReports", sizeLimit)
-      .via(readExprDetails(expressionDetailsDao))
-      .map { case (r, e) => toExprDetails(r, e) }
+      .flatMapConcat { r =>
+        Source
+          .single(r)
+          .via(readExprDetails(expressionDetailsDao))
+          .map(e => (r, e))
+      }
+      .flatMapConcat {
+        case (r, e) =>
+          Source
+            .single(r)
+            .via(lookupScalingPolicy(scalingPolicies))
+            .map(s => (r, s, e))
+      }
+      .map { case (r, s, e) => toExprDetails(r, s, e) }
       .via(saveExprDetails(expressionDetailsDao))
       .viaMat(KillSwitches.single)(Keep.both)
       .toMat(Sink.ignore)(Keep.left)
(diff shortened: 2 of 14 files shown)


iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ScalingPoliciesDao.scala
```diff
@@ -0,0 +1,240 @@
...
+
+          result match {
+            case Failure(e) => logger.error("Request to Edda failed", e)
+            case _          =>
+          }
+
+          result
+      }
+      .collect { case Success(r) => r }
+      .filter(_.status == StatusCodes.OK)
```
brharrington: You should process or dispose of the content for failures.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ScalingPoliciesDao.scala
```diff
@@ -0,0 +1,240 @@
...
+        .map(makeEddaUri(_, cwAlarmsUri))
+        .via(doGet())
+        .map(Json.decode[List[CwAlarm]](_))
+
+      broadcast ~> ec2 ~> zip.in0
+      broadcast ~> titus ~> zip.in1
+      broadcast ~> alarms ~> zip.in2
+
+      FlowShape(broadcast.in, zip.out)
+    })
```
brharrington: It took me a while to follow this graph, if I'm following it correctly the main reason for zipping is because the alarms data is needed for creating the final scaling policy objects both for EC2 and Titus. One alternative that might be easier to understand would be to fetch the alarms and then just flat map the EC2 and Titus scaling policy results, something like:

```scala
Source.single(eddaEndpoint)
  .map(makeEddaUri(_, cwAlarmsUri))
  .via(doGet())
  .map(Json.decode[List[CwAlarm]](_))
  .flatMapConcat { alarms =>
    val ec2 = Source.single(eddaEndpoint)
      ...
      .map(policies => makeEc2ScalingPolicies(policies, alarms))
    val titus = Source.single(eddaEndpoint)
      ...
      .map(policies => makeTitusScalingPolicies(policies, alarms))
    Source.combine(ec2, titus)(Merge(_))
  }
```

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ScalingPolicies.scala
```diff
@@ -0,0 +1,89 @@
...
+  override def receive: Receive = {
+    case GetScalingPolicy(metricInfo) => respond(getScalingPolicy(metricInfo))
+    case ClearCache                   => clearCache()
+    case AddScalingPolicy(policies)   => scalingPolicies = scalingPolicies ++ policies
+    case GetCache                     => sender() ! scalingPolicies
+    case message                      => throw new RuntimeException(s"Unknown message ${message}")
+  }
+
+  def startPeriodicTimer(): Unit = {
+    timers.startPeriodicTimer(NotUsed, ClearCache, eddaCacheTtl)
```
brharrington: This is a bit odd for a cache TTL. It isn't actually the TTL for an entry, but the max time until the entire cache is cleared. Maybe it should just receive a messages to collect the set of Edda endpoints and then use the timer to refresh the data from those endpoints. It also makes the cache age easy to track for alerting.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/MarkerService.scala
```diff
@@ -120,18 +168,54 @@ object MarkerServiceImpl extends StrictLogging {
         )
       )
 
-    val noScalingPolicyFoundEvent = Map.empty[String, Long]
+    val noScalingPolicyFoundEvent =
+      if (scalingPolicyStatus.unknown || scalingPolicyStatus.scalingPolicy.isDefined) {
```
brharrington: What is the difference between the `unknown` setting in `ScalingPolicyStatus` and the option being non empty? At first glance it seems like if there is a scaling policy then `unknown == false`.  Is there a case where these differ?

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/MarkerService.scala
```diff
@@ -120,18 +168,54 @@ object MarkerServiceImpl extends StrictLogging {
...
+  def addOrMake[T](
+    item: Option[T],
+    prev: Option[ExpressionDetails],
+    getList: ExpressionDetails => List[T]
+  ): List[T] = {
+    item
+      .map { i =>
+        prev
+          .map { p =>
+            val l = getList(p)
```
brharrington: Consider using a different variable name, a single `l` is easy to confuse with `1`.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/MarkerService.scala
```diff
@@ -120,18 +168,54 @@ object MarkerServiceImpl extends StrictLogging {
...
-      None
+      addOrMake(scalingPolicyStatus.scalingPolicy, prevExprDetails, _.scalingPolicies)
     )
   }
 
+  def addOrMake[T](
+    item: Option[T],
+    prev: Option[ExpressionDetails],
+    getList: ExpressionDetails => List[T]
+  ): List[T] = {
```
brharrington: I think this method could be simplified to something like:

```scala
val prevItems = prev.fold(List.empty[T])(getList)
item.foldLeft(prevItems) { (items, i) =>
  (i :: items).distinct
}
```

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ScalingPoliciesDao.scala
```diff
@@ -0,0 +1,240 @@
...
+          result
+      }
+      .collect { case Success(r) => r }
+      .filter(_.status == StatusCodes.OK)
+      .map(Gzip.decodeMessage(_))
+      .mapAsync[String](1)(r => Unmarshal(r.entity).to[String])
+  }
+
+  def makeEddaUri(eddaEndpoint: EddaEndpoint, uriPattern: String): Uri = {
+    Uri(uriPattern.format(eddaEndpoint.account, eddaEndpoint.region, "test"))
```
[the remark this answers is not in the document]
narayaruna: Typo

iep-lwc-fwding-admin/src/main/resources/application.conf
```diff
@@ -18,5 +26,10 @@ blocking-dispatcher {
 iep.lwc.fwding-admin {
   age-limit = 10m
   queue-size-limit = 10000
+  eddaCacheTtl = 30m
   cw-expr-uri = "http://localhost:7102/api/v2/cloudwatch-forwarding/clusters/%s"
+
+  cw-alarms-uri = "http://localhost:7103/api/v2/aws/alarms;namespace=NFLX/EPIC;_expand:(alarmName,metricName,dimensions:(name,value))"
+  ec2-policies-uri = "http://localhost:7103/api/v2/aws/scalingPolicies;_expand:(policyName,alarms:(alarmName))"
+  titus-policies-uri = "http://localhost:7103/api/v3/netflix/titusscalingPolicies;_expand:(jobId,id,scalingPolicy:(targetPolicyDescriptor:(customizedMetricSpecification)))"
```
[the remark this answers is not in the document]
narayaruna: Should not be. I ran it against prod/us-east-1 and I see 995 policies in the result.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ScalingPoliciesDao.scala
```diff
@@ -0,0 +1,240 @@
...
+
+  private val ec2PoliciesUri = config.getString("iep.lwc.fwding-admin.ec2-policies-uri")
+  private val cwAlarmsUri = config.getString("iep.lwc.fwding-admin.cw-alarms-uri")
+  private val titusPoliciesUri = config.getString("iep.lwc.fwding-admin.titus-policies-uri")
+
+  protected val client = Http().superPool[AccessLogger]()
+
+  override def getScalingPolicies(
+    eddaEndpoint: EddaEndpoint
+  ): Future[List[ScalingPolicy]] = {
```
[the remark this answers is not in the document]
narayaruna: The ScalingPolicies actor abstracts looking up scaling policies from Edda and caching the result. This actor responds to an ask either from a cached result or by running the Flow to lookup Edda. Right now the flow is materialized in the DAO itself. I could defer running the flow in the [actor](https://github.com/narayaruna/iep-apps/blob/master/iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ScalingPolicies.scala#L38) but I think the current approach might be enough unless I am missing something. Will discuss this with you after I am back.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/MarkerService.scala
narayaruna: That flag is to differentiate option being empty because scaling policy is not created or there was a failure while looking up.
unknown=true,option=empty : Failure in looking up scaling policy
unknown=false,option=empty : No scaling policy found
unknown=false,option=non-empty : unknown is not relevant because scaling policy is found. However it will be set to false to keep the flag consistent.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ScalingPolicies.scala
narayaruna: Yes the timer is for all the entries. I see one advantage in creating TTL per endpoint. The requests to Edda will be distributed.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ScalingPoliciesDao.scala
narayaruna: That's correct. Also I wanted to send all the requests concurrently.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ScalingPolicies.scala
narayaruna: Refreshing the cache asynchronously would speed up reports processing. I will go ahead implement this.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/ScalingPoliciesDao.scala
narayaruna: I changed the implementation to look like the one you suggested except that I am making Edda calls one after other, in order to be able to produce output from the stream only if all the Edda calls succeed.

iep-lwc-fwding-admin/src/main/scala/com/netflix/iep/lwc/fwd/admin/MarkerService.scala
narayaruna: Awesome. I went ahead a replaced with this implementation. Right now, I don't keep track of last reported timestamps for these entities. Because the apps would produce metrics for new Asgs after a new code deploy, these lists would grow and to shrink them the timestamps will be necessary. Long story short, this implementation might change when I work on implementing the shrink logic. 
For now, these two lists would grow and we can remove all entries from the db and the Marker service would recreate the entries when the new reports are processed.

PR #135, opened 2019-04-29
DynamoDB Capacity Updates

copperlight:
The required DynamoDB capacity can differ based on account and region,
depending on the number and size of ASGs being tracked. This change
provides the ability to specify the desired read and write capacity per
account and region, with a default value as the fallback.

The timestamp update and deactivate ASG methods do not change the data
payload in DynamoDB, so the condition expressions have been updated to
check the data that is changing. the data payloads can be up to 400KB in
size, while a Boolean and a Long are just a few bytes. this should help
save some DynamoDB write capacity. keeping the conditional checks in
place will provide stability for the items, when there are multiple
writers during a deployment.

### DIFF
--- a/atlas-slotting/src/main/resources/application.conf
+++ b/atlas-slotting/src/main/resources/application.conf
@@ -19,8 +19,16 @@ aws {
 
   dynamodb {
     table-name = "atlas_slotting-"${netflix.iep.env.stack}
-    read-capacity = 15
-    write-capacity = 15
+
+    read-capacity {
+      default = 15
+      # add $env.$region scoped values here
+    }
+
+    write-capacity {
+      default = 15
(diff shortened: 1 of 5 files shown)


atlas-slotting/src/scripts/compare-data.py
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env python3
```
brharrington: Ideally this could be done as a separate PR since it isn't related to the capacity updater:

https://github.com/Netflix-Skunkworks/iep-apps/blob/master/CONTRIBUTING.md#contributing-code

copperlight: Will do.

Issue #151, opened 2019-06-13
update ses monitor to latest alpakka


brharrington:
The SES monitor needs to get updated to the 1.x release of alpakka which will require a bit of work because it uses the 2.x AWS SDK. Or ideally just use the AWS 2.x SDK directly to avoid the extra dependency.

This will be needed to get the project to build with scala 2.13.

zimmermatt:
Fixed via #152.

I'll take a look at what we gain and lose by moving to use the AWS SDK directly. For now, it was relatively quick and straight-forward to update from the 1.0 milestone to the latest GA version.

PR #152, opened 2019-06-14
Update akka-stream-alpakka-sqs to 1.0.2

zimmermatt:
alpakka-sqs 1.0.+ uses AWS SDK v2, so this update required adapting to those
changes as well.

Issue #151

### DIFF
--- a/iep-ses-monitor/src/main/scala/com/netflix/iep/ses/AppModule.scala
+++ b/iep-ses-monitor/src/main/scala/com/netflix/iep/ses/AppModule.scala
@@ -15,15 +15,14 @@
  */
 package com.netflix.iep.ses
 
-import com.amazonaws.PredefinedClientConfigurations
-import com.amazonaws.auth.DefaultAWSCredentialsProviderChain
-import com.amazonaws.services.sqs.AmazonSQSAsync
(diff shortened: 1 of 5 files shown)


iep-ses-monitor/src/main/scala/com/netflix/iep/ses/AppModule.scala
```diff
@@ -32,13 +31,11 @@ class AppModule extends AbstractModule {
...
   @Provides
-  def provideAwsSqsAsyncClient(): AmazonSQSAsync = {
-    // TODO add to AwsClientFactory?
-    AmazonSQSAsyncClient
-      .asyncBuilder()
-      .withRegion(NetflixEnvironment.region())
-      .withCredentials(new DefaultAWSCredentialsProviderChain)
-      .withClientConfiguration(PredefinedClientConfigurations.defaultConfig().withGzip(true))
+  def provideAwsSqsAsyncClient(): SqsAsyncClient = {
+    SqsAsyncClient
```
brharrington: Ideally this would be done with `AwsClientFactory` from `iep-module-aws2` so it can be configured in a consistent way across our apps.

zimmermatt: Good point. Will do.

PR #153, opened 2019-06-14
Migrate to iep.aws2.AtlasClientFactory

zimmermatt:
No async client was available in the factory for the AWS 1.x SDK and,
since the SES Monitor usage was a one-off, I didn't add it. Now that
alpakka-sqs uses the AWS 2.x SDK and there is a client factory for that
available, I'm moving the SES Monitor to leverage it.

### DIFF
--- a/iep-ses-monitor/src/main/scala/com/netflix/iep/ses/AppModule.scala
+++ b/iep-ses-monitor/src/main/scala/com/netflix/iep/ses/AppModule.scala
@@ -17,6 +17,7 @@ package com.netflix.iep.ses
 
 import com.google.inject.AbstractModule
 import com.google.inject.Provides
+import com.google.inject.Singleton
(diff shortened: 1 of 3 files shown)


iep-ses-monitor/src/main/scala/com/netflix/iep/ses/AppModule.scala
```diff
@@ -31,12 +29,8 @@ class AppModule extends AbstractModule {
   }
 
   @Provides
```
brharrington: Not specifically related to this change, but this should probably be a singleton so the client instance will get reused rather than creating a new client each time it is injected. That said, it shouldn't be much of an issue for this app right now because it will only get injected once.

zimmermatt: noted. will update.

PR #173, opened 2019-10-29
allow streaming multiple urls together

jfz:
add a POST api endpoint to allow streaming multiple urls together

### DIFF
--- a/atlas-stream/src/main/scala/com/netflix/atlas/stream/StreamApi.scala
+++ b/atlas-stream/src/main/scala/com/netflix/atlas/stream/StreamApi.scala
@@ -17,22 +17,28 @@ package com.netflix.atlas.stream
 
 import java.time.Duration
 
-import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.model.HttpEntity
 import akka.http.scaladsl.model.HttpResponse
 import akka.http.scaladsl.model.MediaTypes
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.model.headers.Connection
+import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.Route
 import akka.stream.ThrottleMode
+import akka.stream.scaladsl.Flow
 import akka.stream.scaladsl.Source
 import akka.util.ByteString
 import com.netflix.atlas.akka.CustomDirectives._
 import com.netflix.atlas.akka.DiagnosticMessage
 import com.netflix.atlas.akka.WebApi
 import com.netflix.atlas.eval.stream.Evaluator
+import com.netflix.atlas.eval.stream.Evaluator.DataSource
+import com.netflix.atlas.eval.stream.Evaluator.DataSources
+import com.netflix.atlas.eval.stream.Evaluator.MessageEnvelope
 import com.netflix.atlas.json.Json
 
+import scala.concurrent.duration._
+import scala.jdk.CollectionConverters._
 import scala.util.Failure
 import scala.util.Success
 import scala.util.Try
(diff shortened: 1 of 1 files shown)


atlas-stream/src/main/scala/com/netflix/atlas/stream/StreamApi.scala
```diff
@@ -48,21 +54,43 @@ class StreamApi(evaluator: Evaluator) extends WebApi {
...
+
+          //repeat and throttle to keep the source alive, actual rate of stream is decided by step
+          val src = Source
+            .repeat(dataSources)
+            .throttle(1, 15.seconds, 1, ThrottleMode.Shaping)
+            .via(Flow.fromProcessor(evaluator.createStreamsProcessor))
+            .map { messageEnvelope =>
+              prefix ++ ByteString(Json.encode[MessageEnvelope](messageEnvelope)) ++ suffix
+            }
+            .merge(getHeartBeatSource)
```
brharrington: The duplication could be reduced with a helper method that takes the `DataSources` object and returns the `HttpResponse`.

atlas-stream/src/main/scala/com/netflix/atlas/stream/StreamApi.scala
```diff
@@ -86,4 +114,10 @@ class StreamApi(evaluator: Evaluator) extends WebApi {
       }
     }
   }
+
+  private def getHeartBeatSource = {
```
brharrington: Heartbeat is one word, seems like `s/HeartBeat/Heartbeat/` would be more appropriate.

atlas-stream/src/main/scala/com/netflix/atlas/stream/StreamApi.scala
jfz: done, and I'm not extract "merge" and "complete" to the new method since they belong to the flow building logic

atlas-stream/src/main/scala/com/netflix/atlas/stream/StreamApi.scala
jfz: agree

atlas-stream/src/main/scala/com/netflix/atlas/stream/StreamApi.scala
```diff
@@ -48,24 +55,42 @@ class StreamApi(evaluator: Evaluator) extends WebApi {
...
-
           val q = uri.rawQueryString.getOrElse("")
           val atlasUri = s"$path?$q"
 
-          val heartbeatSrc = Source
-            .repeat(heartbeat)
-            .throttle(1, 5.seconds, 1, ThrottleMode.Shaping)
-
           val src = Source
             .fromPublisher(evaluator.createPublisher(atlasUri))
```
brharrington: I guess what I was thinking is that for the `GET /stream/$uri` case it could be refactored to create a `DataSources` object and use the processor rather than the publisher:

```scala
val dataSources = DataSources.of(new DataSource("_", atlasUri))
complete(createResponse(dataSources))
```

The `POST /stream` case already creates a `DataSources` object, once that is done it would similarly be:

```
complete(createResponse(dataSources))
```

atlas-stream/src/main/scala/com/netflix/atlas/stream/StreamApi.scala
```diff
@@ -48,24 +54,26 @@ class StreamApi(evaluator: Evaluator) extends WebApi {
...
-              prefix ++ ByteString(obj.toJson) ++ suffix
+          val dataSources = DataSources.of(new DataSource("_", atlasUri))
+          processStream(dataSources)
+        }
+      }
+    } ~
+    endpointPath("stream") {
+      post {
+        parseEntity(json[List[DataSource]]) { dsList =>
+          val dsListWithDefaultStep = dsList.map { ds =>
```
brharrington: Ideally this step wouldn't be necessary, I created Netflix/atlas#1101 to have it handled better internally to `DataSource`

PR #186, opened 2020-02-06
atlas-stream: single evaluator and websocket API

jfz:
1. Add a WebSocket endpoint
2. Create a singleton background stream which merges DataSources from all active WebSocket sessions to one, call LWC API's, and routes data back to WebSocket sessions.

### DIFF
--- a/atlas-stream/src/main/scala/com/netflix/atlas/stream/DataSourceInput.scala
+++ b/atlas-stream/src/main/scala/com/netflix/atlas/stream/DataSourceInput.scala
@@ -1,0 +1,121 @@
+/*
+ * Copyright 2014-2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.atlas.stream
+
+import com.netflix.atlas.eval.stream.Evaluator.DataSource
+import com.netflix.atlas.eval.stream.Evaluator.DataSources
+import com.netflix.atlas.json.Json
+import com.netflix.atlas.stream.DataSourceInput.IdAndError
+import com.typesafe.scalalogging.StrictLogging
+
+import scala.collection.mutable
+import scala.jdk.CollectionConverters._
+import scala.util.Failure
+import scala.util.Success
+import scala.util.Try
+
+// Parse and validation DataSource's
+class DataSourceInput private (validateFunc: DataSource => Unit) extends StrictLogging {
+
+  private val errorMap = mutable.Map[String, mutable.Set[String]]()
+  private var dataSourceList: List[DataSource] = List.empty[DataSource]
+
+  def this(input: String, validateFunc: DataSource => Unit) {
+    this(validateFunc)
+    parse(input)
+    validate()
+  }
+
+  def this(dataSourceList: List[DataSource], validateFunc: DataSource => Unit) {
+    this(validateFunc)
+    this.dataSourceList = dataSourceList
+    validate()
+  }
+
+  def isValid: Boolean = errorMap.isEmpty
+
+  def dataSources: DataSources = {
+    if (isValid) {
+      new DataSources(dataSourceList.toSet.asJava)
+    } else {
+      null
+    }
+  }
+
+  // Validation errors as a map, key is the DataSource id, id "_" means it's a general error
+  def errors: List[IdAndError] = {
+    errorMap
+      .map {
+        case (id, errorList) => IdAndError(id, errorList.mkString("; "))
+      }
+      .toList
+      .sortBy(_.id)
+  }
+
+  private def parse(input: String): Unit = {
+    // Parse json
+    try {
+      dataSourceList = Json.decode[List[DataSource]](input)
+    } catch {
+      case e: Exception =>
+        addError("_", s"failed to parse input as List[DataSource]: ${e.getMessage}")
+    }
+  }
+
+  private def validate(): Unit = {
+    // Validate each DataSource
+    val visitedIds = mutable.Set[String]()
+    dataSourceList.foreach(ds => {
+      val id = ds.getId
+
+      // Validate id
+      if (id == null) {
+        addError(id, "id cannot be null")
+      } else if (id.isEmpty) {
+        addError(id, "id cannot be empty")
+      } else {
+        if (visitedIds.contains(id)) {
+          addError(id, "id cannot be duplicated")
+        } else {
+          visitedIds.add(id)
+        }
+      }
+
+      // Validate uri
+      Try(validateFunc(ds)) match {
+        case Success(_) =>
+        case Failure(e) => addError(id, s"invalid uri: ${e.getMessage}")
+      }
+    })
+  }
+
+  private def addError(id: String, value: String): Unit = {
+    // Normalize id for null and empty
+    val idNormalized =
+      if (id == null) {
+        "<null_id>"
+      } else if (id.isEmpty) {
+        "<empty_id>"
+      } else {
+        id
+      }
+    errorMap.getOrElseUpdate(idNormalized, mutable.Set[String]()) += value
+  }
+}
+
+object DataSourceInput {
+  case class IdAndError(id: String, error: String)
+}
--- a/atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalFlow.scala
+++ b/atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalFlow.scala
@@ -87,9 +87,10 @@ private[stream] class EvalFlow(
 
       override def onPush(): Unit = {
         DataSourceValidator.validate(grab(in), validateFunc) match {
-          case Left(dss) => evalService.updateDataSources(streamId, dss)
-          case Right(errors) =>
+          case Left(errors) =>
             queue.offer(new MessageEnvelope("_", DiagnosticMessage.error(Json.encode(errors))))
+          case Right(dss) =>
+            evalService.updateDataSources(streamId, dss)
         }
         pull(in)
       }
--- a/atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalFlow.scala
+++ b/atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalFlow.scala
@@ -31,20 +31,11 @@ import akka.stream.stage.InHandler
 import akka.stream.stage.OutHandler
 import com.netflix.atlas.akka.DiagnosticMessage
 import com.netflix.atlas.akka.StreamOps
-import com.netflix.atlas.eval.stream.Evaluator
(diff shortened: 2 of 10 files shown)


atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalService.scala
```diff
@@ -0,0 +1,208 @@
...
+
+  }
+
+  def updateDataSources(streamId: String, dataSources: DataSources): Unit = {
+    val streamInfo = registrations.get(streamId)
+    if (streamInfo == null) {
+      throw new IllegalStateException(s"stream with id '$streamId' has not been registered")
+    }
+
+    dataSources.getSources.asScala.map(f = ds => new DataSource(ds.getId, ds.getStep, ds.getUri))
```
brharrington: This line doesn't appear to be doing anything. The return value is never used and from what I can tell it doesn't have a side effect.

atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalService.scala
```diff
@@ -0,0 +1,208 @@
...
+          )
+          .asJava
+      )
+    )
+  }
+
+  private def getCurrentDataSources: DataSources = {
+    val dsSet = new util.HashSet[DataSource]
+    registrations.values.asScala
+      .foreach(streamInfo => streamInfo.dataSources.map(dss => dsSet.addAll(dss.getSources)))
```
brharrington: You could just use `forEach` on the Java collection:

```java
registrations.values.forEach(_.dataSources.foreach(dss => dsSet.addAll(dss.getSources)))
```

More idiomatic Scala would be something like:

```scala
val dsSet = registrations
      .values
      .asScala
      .flatMap(_.dataSources)
      .toSet
      .asJava
```

atlas-stream/src/main/scala/com/netflix/atlas/stream/StreamApi.scala
```diff
@@ -43,14 +49,19 @@ import scala.util.Failure
...
-class StreamApi(evaluator: Evaluator) extends WebApi {
+class StreamApi(evaluator: Evaluator, evalService: EvalService) extends WebApi {
 
   private val prefix = ByteString("data: ")
   private val suffix = ByteString("\r\n\r\n")
 
   private val heartbeat = ByteString(s"""data: {"type":"heartbeat"}\r\n\r\n""")
 
   def routes: Route = {
+    endpointPath("streamws") {
```
brharrington: We might be able to chain this with the existing endpoint. Might be worth thinking a bit about what convention we want for websocket usage across APIs. In particular ones that other users might hit and are not just internal details. Ideally we can be consistent across our services.

atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalFlow.scala
```diff
@@ -0,0 +1,205 @@
...
+          .merge(heartbeatSrc, true) // Eager complete: heartbeat never completes
+      }
+
+      override def onPush(): Unit = {
+        val parser = DataSourcesParser(grab(in), evaluator.validate)
+
+        if (parser.isValid) {
+          evalService.updateDataSources(streamId, parser.dataSources)
+        } else {
+          // All or None, ignore all if any validation issue
```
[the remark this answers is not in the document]
jfz: The subscription update is replacing previous one completely, so I think "all or none" would reduce confusion compared to "partial update".  And the validation logic should be consistent between this API and "validation" API, I'm thinking of using the same Parser for validation API too.

atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalFlow.scala
```diff
@@ -0,0 +1,205 @@
...
+          queue.offer(
+            new MessageEnvelope(
+              "_",
+              DiagnosticMessage.error(
+                s"Invalid input: ${parser.errors.mkString("{", ", ", "}")}"
```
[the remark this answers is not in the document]
jfz: Same reason for "all or none",  there should be a clear indication of result of the "subscription", I'll put more thoughts on this.

build.sbt
```diff
@@ -73,7 +73,8 @@ lazy val `atlas-stream` = project
   .configure(BuildSettings.profile)
   .settings(libraryDependencies ++= Seq(
     Dependencies.atlasModuleAkka,
-    Dependencies.atlasModuleEval,
+//    Dependencies.atlasModuleEval,
+    "com.netflix.atlas_v1" %% "atlas-module-eval" % "1.7.0-SNAPSHOT",
```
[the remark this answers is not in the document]
jfz: that apply to all other apps, is that ok?

atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalService.scala
```diff
@@ -0,0 +1,208 @@
...
+            new MessageEnvelope(envelope.getId.substring(index + 1), envelope.getMessage)
+          )
+        } else {
+          logger.debug(s"discarding message without handler: $envelope")
+        }
+      } else {
+        logger.debug(s"discarding message without streamId: $envelope")
+      }
+    } catch {
+      case t: Throwable => {
```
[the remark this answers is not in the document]
jfz: agreed

atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalService.scala
```diff
@@ -0,0 +1,208 @@
...
+        if (info != null) {
+          info.handler.offer(
+            // Remove prefix
+            new MessageEnvelope(envelope.getId.substring(index + 1), envelope.getMessage)
+          )
+        } else {
+          logger.debug(s"discarding message without handler: $envelope")
+        }
+      } else {
+        logger.debug(s"discarding message without streamId: $envelope")
```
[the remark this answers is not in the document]
jfz: right, I guess I could remove this, and it can be captured by Exception handling in worst case.

atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalFlow.scala
```diff
@@ -0,0 +1,205 @@
...
+      val visitedIds = mutable.Set[String]()
+      dataSourceList.foreach(ds => {
+        val id = ds.getId
+
+        // Validate id
+        if (id == null || id.isEmpty) {
+          addError("_", "id cannot be empty")
+        } else {
+          if (visitedIds.contains(id)) {
+            addError(id, "id cannot be duplicated")
```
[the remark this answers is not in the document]
jfz: By enforcing uniq id here we can avoid users running into issue and come back to us for help, it can be part of API spec.

build.sbt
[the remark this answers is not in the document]
brharrington: Internally we run most with the snapshot version anyway. We can always do an atlas release before the final merge and update to the latest.

atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalFlow.scala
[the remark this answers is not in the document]
brharrington: Fair enough, we should probably think a bit about whether we want this to match the behavior of using the eval processor directly. Should it reject duplicate ids as well?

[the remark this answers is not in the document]
jfz: I think so, better to fail fast and loud for the same reason.

atlas-stream/src/main/scala/com/netflix/atlas/stream/StreamApi.scala
jfz: ya same end point "/stream" works given different http method / protocol

atlas-stream/src/main/scala/com/netflix/atlas/stream/DataSourceInput.scala
```diff
@@ -0,0 +1,121 @@
...
+      } else {
+        if (visitedIds.contains(id)) {
+          addError(id, "id cannot be duplicated")
+        } else {
+          visitedIds.add(id)
+        }
+      }
+
+      // Validate uri
+      Try(validateFunc(ds)) match {
```
brharrington: This is using exceptions for control flow, why not just have the function return the validation result?

atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalFlow.scala
```diff
@@ -0,0 +1,130 @@
...
+        val heartbeatSrc = Source
+          .repeat(new MessageEnvelope("_", DiagnosticMessage.info("heartbeat")))
+          .throttle(1, 5.seconds, 1, ThrottleMode.Shaping)
+        Source
+          .fromPublisher(pub)
+          .merge(heartbeatSrc, eagerComplete = true) // Eager complete: heartbeat never completes
+      }
+
+      override def onPush(): Unit = {
+        val dataSourceInput = new DataSourceInput(grab(in), validateFunc)
```
brharrington: Do we need to create the temporary `DataSourceInput` object? It seems like we could just have a function that returns `Either` or some custom type to indicate the result.

atlas-stream/src/main/scala/com/netflix/atlas/stream/StreamApi.scala
```diff
@@ -92,9 +103,43 @@ class StreamApi(evaluator: Evaluator) extends WebApi {
...
     }
   }
 
+  private def createHandler(): Flow[Message, Message, Any] = {
+    Flow[Message]
+      .flatMapConcat {
+        case msg: TextMessage =>
+          msg.textStream.fold("")(_ + _)
+        case msg: BinaryMessage =>
+          msg.dataStream.fold(ByteString.empty)(_ ++ _).map(_.decodeString(StandardCharsets.UTF_8))
```
brharrington: Since the text message format is UTF-8, not sure why someone would send UTF-8 as a binary message. I wonder if we should allow a binary format that is more efficient or maybe reject binary messages for now and leave it open for future improvement.

atlas-stream/src/main/scala/com/netflix/atlas/stream/DataSourceInput.scala
jfz: Evaluator.validate is a void method, don't want have another layer of indirect

atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalFlow.scala
jfz: Changed to return Either, like that style - way more neat! 😄

atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalFlow.scala
```diff
@@ -86,15 +86,10 @@ private[stream] class EvalFlow(
...
-        if (dataSourceInput.isValid) {
-          evalService.updateDataSources(streamId, dataSourceInput.dataSources)
-        } else {
-          // All or None, ignore all if any validation issue
-          queue.offer(
-            new MessageEnvelope("_", DiagnosticMessage.error(Json.encode(dataSourceInput.errors)))
-          )
+        DataSourceValidator.validate(grab(in), validateFunc) match {
+          case Left(dss) => evalService.updateDataSources(streamId, dss)
+          case Right(errors) =>
```
brharrington: https://www.scala-lang.org/api/2.12.0/scala/util/Either.html

> Convention dictates that Left is used for failure and Right is used for success.

jfz: Thanks, I'll make this change and merge

PR #187, opened 2020-02-18
stop all sessions on EvalService restart

jfz:
Currently, when EvalService restarts for any error, it completes queues for all active sessions, and because EvalFlow pushes out a new Source continuously whenever previous one completes, and the initial heartbeat is always available immediately, which may cause super high cpu/network consumption at both client and server side.

The fix is to have EvalFlow only push out Source once and simply completeStage if it ever gets pulled again due to completion of previous source. This means all active session will be closed on EvalService restart, which I think should be the right thing to do.

Also add some comments about why not doing synchronization between `cleanup` and `register` on EvalService restart.

### DIFF
--- a/atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalService.scala
+++ b/atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalService.scala
@@ -85,6 +85,14 @@ class EvalService @Inject()(
     cleanup()
   }
 
+  /**
+    * For now, not doing synchronization with `register`, considering below scenarios:
+    * 1. registered but not completed here, then cleared from Map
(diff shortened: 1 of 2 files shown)


atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalService.scala
```diff
@@ -69,6 +69,12 @@ class EvalService @Inject()(
 
     dssSource
       .via(Flow.fromProcessor(() => evaluator.createStreamsProcessor()))
+      .map(msg => {
```
brharrington: This looks like it was for testing and should be removed.

jfz: yes, fixed

PR #194, opened 2020-02-26
atlas-stream: add a metric to track number of DataSource's

jfz:
atlas-stream: add a metric to track number of DataSource's in EvalSerice


atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalService.scala
```diff
@@ -180,6 +181,7 @@ class EvalService @Inject()(
       .flatMap(_.getSources.asScala)
       .toSet
       .asJava
+    numDataSourceCounter.increment(dsSet.size())
```
brharrington: I think a gauge or distribution summary would be a better option here. The counter would measure the rate per second for the number of data sources returned by this method and that will be influenced by how frequently it is called. A gauge or distribution summary would allow us to see the actual size.

jfz: make sense, changed to DistributionSummary

PR #201, opened 2020-03-04
atlas-stream: configurable queue size; grep friendly log message

jfz:
atlas-stream: configurable queue size; grep friendly log message

### DIFF
--- a/atlas-stream/src/main/resources/application.conf
+++ b/atlas-stream/src/main/resources/application.conf
@@ -6,3 +6,5 @@ atlas.akka {
     "com.netflix.atlas.stream.StreamApi"
   ]
 }
+
+atlas.stream.evalService.queueSize = 10000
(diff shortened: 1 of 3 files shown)


atlas-stream/src/main/resources/application.conf
```diff
@@ -5,4 +5,6 @@ atlas.akka {
     "com.netflix.atlas.akka.HealthcheckApi",
     "com.netflix.atlas.stream.StreamApi"
   ]
-}
\ No newline at end of file
+}
+
+atlas.stream.evalService.queueSize = 10000
```
brharrington: It is more idiomatic for the typesafe config files to use dashes rather than camel case.

jfz: thanks, changed

PR #206, opened 2020-03-24
atlas-stream: add limit for max-datasources-per-session

jfz:
Limit number of DataSources per session. 
Refactored DataSourceValidator to avoid passing extra parameters everywhere.

### DIFF
--- a/atlas-stream/src/main/resources/application.conf
+++ b/atlas-stream/src/main/resources/application.conf
@@ -7,4 +7,7 @@ atlas.akka {
   ]
 }
 
-atlas.stream.eval-service.queue-size = 10000
+atlas.stream {
+  eval-service.queue-size = 10000
+  max-datasources-per-session = 100
+}
--- a/atlas-stream/src/main/scala/com/netflix/atlas/stream/DataSourceValidator.scala
+++ b/atlas-stream/src/main/scala/com/netflix/atlas/stream/DataSourceValidator.scala
@@ -26,12 +26,9 @@ import scala.util.Success
 import scala.util.Try
 
 // Parse and validation DataSource's
-object DataSourceValidator {
+case class DataSourceValidator(maxDataSourcesPerSession: Int, validateFunc: DataSource => Unit) {
 
-  def validate(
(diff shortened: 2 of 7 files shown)


atlas-stream/src/main/scala/com/netflix/atlas/stream/DataSourceValidator.scala
```diff
@@ -40,43 +37,45 @@ object DataSourceValidator {
...
-        addError(id, "id cannot be null", errorMap)
-      } else if (id.isEmpty) {
-        addError(id, "id cannot be empty", errorMap)
-      } else {
-        if (visitedIds.contains(id)) {
-          addError(id, "id cannot be duplicated", errorMap)
+
+    // Validate size limit first
+    if (dataSourceList.size > maxDataSourcesPerSession) {
+      addError("_", s"number of DataSources cannot exceed $maxDataSourcesPerSession", errorMap)
```
jfz: quickly fail if over limit without extra info to keep it clear

atlas-stream/src/main/scala/com/netflix/atlas/stream/EvalService.scala
```diff
@@ -203,7 +202,7 @@ object EvalService {
     }
 
     def complete(): Unit = {
-      logger.debug(s"queue complete for $id")
+      logger.info(s"queue complete for: $id")
```
jfz: worth log it for troubleshooting, one per session only

atlas-stream/src/main/resources/application.conf
```diff
@@ -7,4 +7,7 @@ atlas.akka {
   ]
 }
 
-atlas.stream.eval-service.queue-size = 10000
\ No newline at end of file
+atlas.stream {
+  eval-service.queue-size = 10000
+  max-datasources-per-session = 100
```
brharrington: Will you also be adding a limit for the max overall?

jfz: yes

PR #209, opened 2020-04-03
atlas-stream: sending some info messages for websocket api

jfz:
atlas-stream: sending some info messages for websocket api


atlas-stream/src/main/scala/com/netflix/atlas/stream/StreamApi.scala
```diff
@@ -67,7 +67,13 @@ class StreamApi(
     endpointPath("stream") {
       extractUpgradeToWebSocket { upgrade =>
         if (evalService.getNumDataSources > maxDataSourcesTotal) {
-          complete(HttpResponse(StatusCodes.ServiceUnavailable, Nil))
+          complete(
+            HttpResponse(
```
brharrington, line 71: Consider using `DiagnosticMessage.error(StatusCodes.ServiceUnavailable, message)` instead. This would keep the response consistent with other errors.

PR #217, opened 2020-05-07
Hourly bucketing to allow late events within limit

jfz:


atlas-persistence/src/main/scala/com/netflix/atlas/persistence/HourlyRollingWriter.scala
```diff
@@ -0,0 +1,136 @@
...
+      prevWriterInfo = null
+    }
+  }
+
+  private def getHourStart(timestamp: Long): Long = {
+    timestamp / msOfOneHour * msOfOneHour
+  }
+
+  private def getFilePathPrefix(hourStart: Long): String = {
+    val dateTime = LocalDateTime.ofInstant(Instant.ofEpochMilli(hourStart), ZoneId.systemDefault())
```
brharrington, line 118: The system default zone could vary, we try to standardize on UTC everywhere except for presenting data to end users.

atlas-persistence/src/main/scala/com/netflix/atlas/persistence/HourlyRollingWriter.scala
```diff
@@ -0,0 +1,136 @@
...
+  dataDir: String,
+  maxLateDuration: Long,
+  writerFactory: String => RollingFileWriter,
+  registry: Registry
+) extends StrictLogging {
+
+  private val hourlyDirFormatter = DateTimeFormatter.ofPattern("yyyyMMdd'T'HH")
+  private val msOfOneHour = 3600000
+
+  private val lateEventCounter = registry.counter("lateEvents")
```
brharrington, line 41: We typically use a short namespace prefix on metric names to make it easier to search for related names. Perhaps `persistence.lateEvents`.

atlas-persistence/src/main/scala/com/netflix/atlas/persistence/HourlyRollingWriter.scala
jfz: changed to UTC

PR #220, opened 2020-05-15
upload locally batched files to S3

jfz:
upload locally batched files to S3


atlas-persistence/src/main/scala/com/netflix/atlas/persistence/FileWatchSource.scala
```diff
@@ -0,0 +1,108 @@
...
+    new GraphStageLogic(shape) with OutHandler {
+
+      private var watchService: WatchService = _
+      private val queue = mutable.Queue[File]()
+
+      setHandler(out, this)
+
+      override def preStart(): Unit = {
+        logger.debug("starting WatchService")
+        checkExistingFiles
```
brharrington, line 52: Should this be called after registering the watch service? In theory there is a race condition where a file could get added between the check and the registration.

jfz: Moved it to after registration, there still can be race condition of file being double counted, but that's better than missing a file.

PR #221, opened 2020-05-19
atlas-persistence: basic error handling

jfz:
atlas-persistence: basic error handling with auto restart, logging and metrics

### DIFF
--- a/atlas-persistence/src/main/scala/com/netflix/atlas/persistence/RollingFileWriter.scala
+++ b/atlas-persistence/src/main/scala/com/netflix/atlas/persistence/RollingFileWriter.scala
@@ -62,7 +62,7 @@ class AvroRollingFileWriter(
 
   private var nextFileSeqId: Long = 0
 
-  private val avroWriteError = registry.counter("persistence.avroWriteError")
+  private val avroWriteErrors = registry.counter("persistence.avroWriteErrors")
 
   override protected def newWriter(): Unit = {
     val newFile = getNextTmpFilePath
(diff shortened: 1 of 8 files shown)


atlas-persistence/src/main/scala/com/netflix/atlas/persistence/RollingFileWriter.scala
```diff
@@ -80,7 +83,14 @@ class AvroRollingFileWriter(
   }
 
   override protected def writeImpl(dp: Datapoint): Unit = {
-    currWriter.append(toAvro(dp))
+    try {
+      currWriter.append(toAvro(dp))
+    } catch {
+      case e: Exception => {
+        avroWriteError.increment()
+        logger.debug(s"error writing to avro file, file=$currFile datapoint=$dp", e)
```
brharrington, line 91: Why debug level instead of warn or error?

atlas-persistence/src/main/scala/com/netflix/atlas/persistence/RollingFileWriter.scala
```diff
@@ -61,6 +62,8 @@ class AvroRollingFileWriter(
 
   private var nextFileSeqId: Long = 0
 
+  private val avroWriteError = registry.counter("persistence.avroWriteError")
```
brharrington, line 65: `avroWriteErrors`?

atlas-persistence/src/main/scala/com/netflix/atlas/persistence/RollingFileWriter.scala
jfz: this is record level, may produce too much log data

atlas-persistence/src/main/scala/com/netflix/atlas/persistence/RollingFileWriter.scala
jfz: sounds good

PR #226, opened 2020-05-26
atlas-persistence: proper handling of file watch and some corner cases

jfz:
A few changes here: 
1. Periodically listing file with dedup for s3 copy service, to replace Java FileWatch service which does miss files occasionally
2. Make sure file seq id is always consecutive for the same run: empty file should not increase seq id
3. Add a timer to trigger rollover timely if writer is idle: very likely to happen while writer is waiting for previous hour data during courtesy period.
4. Will copy .tmp file if inactive for too long
5. S3CopyService will wait for all files to be processed during graceful shutdown


atlas-persistence/src/main/scala/com/netflix/atlas/persistence/S3CopyService.scala
```diff
@@ -48,31 +50,88 @@ class S3CopyService @Inject()(
...
+      }
+    } else {
+      logger.debug(s"Should process: regular file - $f")
+      true
+    }
+  }
+
+  private def isInactive(f: File): Boolean = {
+    try {
+      System.currentTimeMillis > f.lastModified() + maxInactiveMs
```
brharrington, line 100: `lastModified` will return `0` on error, it will not throw an exception.

atlas-persistence/src/main/scala/com/netflix/atlas/persistence/S3CopySink.scala
```diff
@@ -61,10 +62,9 @@ class S3CopySink(
   override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = {
     new GraphStageLogic(shape) with InHandler {
 
-      private var s3Client: S3AsyncClient = _
+      @volatile private var s3Client: S3AsyncClient = _
```
brharrington, line 65: Why does this need to be volatile? My understanding is the framework should guarantee a single thread is calling these methods.

Side note, we typically use [iep-module-aws2](https://github.com/Netflix/iep/tree/master/iep-module-aws2) so we have consistent configuration and metrics for AWS clients.

atlas-persistence/src/main/scala/com/netflix/atlas/persistence/S3CopySink.scala
```diff
@@ -73,21 +73,18 @@ class S3CopySink(
...
       override def onUpstreamFinish(): Unit = {
         super.completeStage()
-        activeFiles.values.foreach(_.shutdown())
+        activeFiles.values.asScala.foreach(_.shutdown())
       }
 
       override def onUpstreamFailure(ex: Throwable): Unit = {
         super.failStage(ex)
-        activeFiles.values.foreach(_.shutdown())
+        activeFiles.values.asScala.foreach(_.shutdown())
```
brharrington, line 87: Should the shutdown happen before forwarding the exception? I wonder if it could throw in some cases and prevent the shutdown.

atlas-persistence/src/main/scala/com/netflix/atlas/persistence/FileUtil.scala
```diff
@@ -15,28 +15,49 @@
...
-      case e: Exception => logger.error(s"failed to delete path $path", e)
+      case e: Exception => logger.error(s"failed to delete path $f", e)
     }
   }
 
-  def move(source: Path, target: Path, logger: Logger): Unit = {
+  def listFiles(f: File): List[File] = {
     try {
-      Files.move(source, target)
+      f.listFiles().toList.filter(_.isFile)
```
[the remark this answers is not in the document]
jfz: right will use Files api instead

atlas-persistence/src/main/scala/com/netflix/atlas/persistence/RollingFileWriter.scala
```diff
@@ -35,7 +34,10 @@ trait RollingFileWriter extends StrictLogging {
       rollOver
       newWriter
     }
-    writeImpl(dp)
+    // Ignore the special datapoint
+    if (RollingFileWriter.RolloverCheckDatapoint ne dp) {
```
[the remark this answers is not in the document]
jfz: RolloverCheckDatapoint is a static constant, so I'm check if reference is the same

atlas-persistence/src/main/scala/com/netflix/atlas/persistence/FileUtil.scala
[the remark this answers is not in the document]
jfz: use Files API instead

atlas-persistence/src/main/scala/com/netflix/atlas/persistence/S3CopySink.scala
jfz: Shutdown first does sounds safer - from what I've seen onUpstreamFailure is usually executed safely and super.failStage(ex) don't really directly throw exception on spot.

atlas-persistence/src/main/scala/com/netflix/atlas/persistence/S3CopySink.scala
jfz: volatile is not needed - I forgot to delete after a experiment; putting TODO for  iep-module-aws2 for now

atlas-persistence/src/main/scala/com/netflix/atlas/persistence/S3CopyService.scala
jfz: updated to check return value instead

PR #229, opened 2020-05-28
atlas-persistence: make file rename atomic to fix early termination

jfz:
File rename was not atomic, it some times cause early termination of waiting during shutdown before all files are processed

### DIFF
--- a/atlas-persistence/src/main/scala/com/netflix/atlas/persistence/S3CopyService.scala
+++ b/atlas-persistence/src/main/scala/com/netflix/atlas/persistence/S3CopyService.scala
@@ -46,7 +46,6 @@ class S3CopyService @Inject()(
 
   private val dataDir = config.getString("atlas.persistence.local-file.data-dir")
 
-  private implicit val ec = scala.concurrent.ExecutionContext.global
   private implicit val mat = ActorMaterializer()
 
   private var killSwitch: KillSwitch = _
@@ -85,6 +84,8 @@ class S3CopyService @Inject()(
   private def waitForCleanup(): Unit = {
     logger.info("Waiting for cleanup")
     val start = System.currentTimeMillis
+    // Wait a bit first as a best effort try in case atomic rename is not supported by FileSystem
+    Thread.sleep(5000)
     while (hasMoreFiles) {
       if (System.currentTimeMillis() > start + cleanupTimeoutMs) {
         logger.error("Cleanup timeout")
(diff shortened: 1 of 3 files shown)


atlas-persistence/src/main/scala/com/netflix/atlas/persistence/S3CopyService.scala
```diff
@@ -85,6 +84,8 @@ class S3CopyService @Inject()(
   private def waitForCleanup(): Unit = {
     logger.info("Waiting for cleanup")
     val start = System.currentTimeMillis
+    // Wait a bit first as a best effort try in case atomic rename is not supported by FileSystem
+    Thread.sleep(5000)
```
brharrington, line 88: Has this been tested with just the change to perform an atomic move to see if it works correctly? It seems more likely to me that the stream in LocalFilePersistService hasn't finished shutting down rather than a race condition with the file system detection.

I think a better option might be to update LocalFilePersistService so it will get the future for the stream and wait for it to complete after calling complete on the queue. That would ensure that LocalFilePersistService has fully shutdown before this class is ever shutdown and should avoid the need for arbitrary sleep statements.

jfz: I did test this, data dir should have the temp file even if File Service is hasn't complete and it would have kept waiting. But actually waitForCleanup sees no file and stopped waiting, while I can find renamed files afterward.

Let me look at the more elegant way of waiting.

PS I tested with this program: without atomic it fails fairly quickly.

public static void flip(String dir) throws IOException {
        boolean flip = true;
        Path a = Paths.get(dir + "/a");
        Path b = Paths.get(dir + "/b");
        while (true) {
            if (flip) {
                System.out.println("a");
                //Files.move(a, b, StandardCopyOption.ATOMIC_MOVE);
                Files.move(a, b);
                flip = false;
            } else {
                System.out.println("### b");
                //Files.move(b, a, StandardCopyOption.ATOMIC_MOVE);
                Files.move(a, b);
                flip = true;
            }
        }
    }

jfz: ATOMIC_MOVE is pretty stable if OS supports it, and waiting for LocalFilePersistService is not sufficient because there may be files being copied, so checking if data dir is empty seems most reliable way.

brharrington: Adding the `ATOMIC_MOVE` is fine, but I would still argue the sleep should be removed and LocalFilePersistService should check the stream future to ensure it has actually completed and called rollOver for the final time before S3CopyService begins to shutdown.

jfz: Agree, let me make it explicit for 2 different waits: write complete and copy complete

PR #230, opened 2020-05-28
atlas-persistence: use future to wait for file writing to complete

jfz:
atlas-persistence: use future to wait for file writing to complete right after writer being closed to avoid arbitrary sleep.

### DIFF
--- a/atlas-persistence/src/main/scala/com/netflix/atlas/persistence/LocalFilePersistService.scala
+++ b/atlas-persistence/src/main/scala/com/netflix/atlas/persistence/LocalFilePersistService.scala
@@ -35,8 +35,8 @@ import javax.inject.Singleton
 class LocalFilePersistService @Inject()(
   val config: Config,
   val registry: Registry,
-  // S3CopyService is actually NOT used by this service, it is here just to guarantee that the
-  // shutdown callback (stopImpl) of this service is invoked before S3CopyService's
+  // S3CopyService as a dependency also guarantees that the shutdown callback (stopImpl) of this
+  // service is invoked before S3CopyService's
   val s3CopyService: S3CopyService,
   implicit val system: ActorSystem
 ) extends AbstractService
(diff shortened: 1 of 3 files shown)


jfz:
> I think it should work, but not exactly what I was thinking. I was thinking in `LocalFilePersistService.stopImpl` you could do something like:
> 
> ```scala
>   override def stopImpl(): Unit = {
>     logger.info("Stopping service")
>     // Close the queue source. This will cause the stream to start shutting down.
>     queue.complete()
>     // Wait for the stream to finish shutting down. Otherwise the method would return
>     // immediatly and cause issues for the S3 copy.
>     Await.result(future, Duration.Inf)
>   }
> ```
> 
> Similar example from test case:
> 
> https://github.com/Netflix/atlas/blob/master/atlas-akka/src/test/scala/com/netflix/atlas/akka/StreamOpsSuite.scala#L135
> 
> The main benefit of this approach is I think it is conceptually a bit easier to follow because `LocalFilePersistService` would be fully stopped before moving forward.

I like the idea of waiting inside LocalFilePersistService, much cleaner, plus there's a flaw with the way I mark complete inside S3CopySink because it can be restarted, let me think about it how to do it best.

jfz:
> I think it should work, but not exactly what I was thinking. I was thinking in `LocalFilePersistService.stopImpl` you could do something like:
> 
> ```scala
>   override def stopImpl(): Unit = {
>     logger.info("Stopping service")
>     // Close the queue source. This will cause the stream to start shutting down.
>     queue.complete()
>     // Wait for the stream to finish shutting down. Otherwise the method would return
>     // immediatly and cause issues for the S3 copy.
>     Await.result(future, Duration.Inf)
>   }
> ```
> 
> Similar example from test case:
> 
> https://github.com/Netflix/atlas/blob/master/atlas-akka/src/test/scala/com/netflix/atlas/akka/StreamOpsSuite.scala#L135
> 
> The main benefit of this approach is I think it is conceptually a bit easier to follow because `LocalFilePersistService` would be fully stopped before moving forward.

jfz:
will implement in a cleaner way

PR #231, opened 2020-05-29
atlas-persistence: wait for write flow complete in shutdown

jfz:
covert RollingFileSink to flow and wait for its complete on shutdown


atlas-persistence/src/main/scala/com/netflix/atlas/persistence/RollingFileFlow.scala
```diff
@@ -74,16 +78,20 @@ class RollingFileSink(
...
-        super.failStage(ex)
         hourlyWriter.close()
+        failStage(ex)
       }
 
-      setHandler(in, this)
+      setHandlers(in, out, this)
+
+      override def onPull(): Unit = {
+        // Nothing to emit
```
brharrington, line 93: I think you could probably move the pull from preStart to this method. I think the normal expectation is that it will pull when there is a request from downstream. Since it never pushes, there wouldn't be another request.

PR #233, opened 2020-06-01
Timeout for s3 call and retry by listing file

jfz:
It seems like the s3 async client call never finishes occasionally (a few files after 2 days), with current flow, those file never get a chance to be re-processed in such cases. So setting a timeout from our end, release from dedup map when timeout and let file listing trigger the re-process.


jfz:
> Do you know what it is blocked on? This seems like a bug either on our side or that we should report upstream to the AWS SDK team.
I'm not sure yet, trying to figure it out. But I think having some control at our end is also necessary.

PR #239, opened 2020-06-25
atlas-persistence: switch to sync s3 api copy

jfz:
Avoid fd leak issue of async S3 client

### DIFF
--- a/atlas-persistence/src/main/resources/application.conf
+++ b/atlas-persistence/src/main/resources/application.conf
@@ -25,7 +25,8 @@ atlas {
       // A ".tmp" file not modified for this duration will be marked as complete (renamed), and it
       // MUST be longer than local-file.max-duration to avoid conflict
       max-inactive-duration = 7m
-      client-timeout = 2m
+      client-timeout = 5m
+      thread-pool-size = 50
     }
   }
 }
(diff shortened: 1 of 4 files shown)


atlas-persistence/src/main/resources/application.conf
```diff
@@ -25,7 +25,8 @@ atlas {
       // A ".tmp" file not modified for this duration will be marked as complete (renamed), and it
       // MUST be longer than local-file.max-duration to avoid conflict
       max-inactive-duration = 7m
-      client-timeout = 2m
+      client-timeout = 5m
+      thread-pool-size = 50
```
brharrington, line 29: That is a pretty large pool size. Typically we use some fraction or multiple of the number of CPUs available on a machine.

atlas-persistence/src/main/scala/com/netflix/atlas/persistence/S3CopySink.scala
```diff
@@ -103,9 +106,10 @@ class S3CopySink(
 
       // TODO use iep/iep-module-aws2
       private def initS3Client(): Unit = {
-        s3Client = S3AsyncClient
+        s3Client = S3Client
           .builder()
           .region(Region.of(region))
+          .httpClientBuilder(ApacheHttpClient.builder())
```
brharrington, line 112: Does this need to be set explicitly? I think the default should be fine here.

atlas-persistence/src/main/resources/application.conf
jfz: This pool is not for event loop but rather waiting for long running sync s3 api calls, so small size may very likely cause congestion.

atlas-persistence/src/main/scala/com/netflix/atlas/persistence/S3CopySink.scala
jfz: ya, let me remove this, and remove explicit dependency too

PR #243, opened 2020-07-28
atlas-persistence: enable data compression

jfz:
Enable data compression to reduce output file size.

### DIFF
--- a/atlas-persistence/src/main/resources/application.conf
+++ b/atlas-persistence/src/main/resources/application.conf
@@ -15,6 +15,10 @@ atlas {
       max-records = 100000
       max-duration = 5m
       max-late-duration = 5m
+      // Compression codec, can be: null, deflate, bzip2 or snappy
+      avro-codec = "null"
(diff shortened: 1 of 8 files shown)


atlas-persistence/src/main/resources/application.conf
```diff
@@ -15,6 +15,10 @@ atlas {
       max-records = 100000
       max-duration = 5m
       max-late-duration = 5m
+      // Compression codec, can be: null, deflate, bzip2 or snappy
+      avro-codec = "null"
```
brharrington, line 19: Seems a bit odd to use the string null. I would prefer something like `none` to avoid confusion with a null reference. Looks like that is was the CodecFactory supports though:

http://avro.apache.org/docs/1.8.1/api/java/org/apache/avro/file/CodecFactory.html#fromString(java.lang.String)

jfz: It did confuse me during dev test :), but it should be caught at start up. And I think keeping it consistent with avro lib's definition is easier to map in mind.

PR #247, opened 2020-08-06
atlas-persistence: use generic avro type to avoid overhead

jfz:
Switch to avro's GenericRecord APIs to avoid overhead. 
Currently a specifc java class is generated for Datapoint and its instances are used for writing to avro file, but that creates quite some overhead because it creates every new instance by reflection(Class.forName), and eventually come down to the same path of generic type and writer.
It also simplify project setup since we don't need the extra step of code gen and schema compilation stuff.

### DIFF
--- a/atlas-persistence/src/main/scala/com/netflix/atlas/persistence/RollingFileWriter.scala
+++ b/atlas-persistence/src/main/scala/com/netflix/atlas/persistence/RollingFileWriter.scala
@@ -23,11 +23,16 @@ import java.nio.file.StandardCopyOption
 import com.netflix.atlas.core.model.Datapoint
 import com.netflix.spectator.api.Registry
 import com.typesafe.scalalogging.StrictLogging
+import org.apache.avro.Schema.Parser
 import org.apache.avro.file.CodecFactory
 import org.apache.avro.file.DataFileWriter
-import org.apache.avro.specific.SpecificDatumWriter
+import org.apache.avro.generic.GenericData
+import org.apache.avro.generic.GenericDatumWriter
+import org.apache.avro.generic.GenericRecord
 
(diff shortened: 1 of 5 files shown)


atlas-persistence/src/main/scala/com/netflix/atlas/persistence/RollingFileWriter.scala
```diff
@@ -177,18 +180,19 @@ class RollingFileWriter(
...
+    record.put("value", dp.value)
+    record
   }
 }
 
 object RollingFileWriter {
   val TmpFileSuffix: String = ".tmp"
   // A special Datapoint used solely for triggering rollover check
   val RolloverCheckDatapoint: Datapoint = Datapoint(Map.empty, 0, 0)
+  val AvroSchema = new Parser().parse(scala.io.Source.fromResource("datapoint.avsc").mkString)
```
brharrington, line 197: You need to ensure the source is closed:

```scala
Using.resource(Source.fromResource("datapoint.avsc"))(_.mkString)
```

project/plugins.sbt
```diff
@@ -8,8 +8,3 @@ addSbtPlugin("org.scalameta"             % "sbt-scalafmt"         % "2.3.1")
 
 // Convenient helpers, not required
 addSbtPlugin("com.timushev.sbt"          % "sbt-updates"          % "0.5.0")
-
-// Avro Java source code generation from schema
-addSbtPlugin("com.cavorite" % "sbt-avro" % "2.0.0").settings
-// avro-compiler version should match actual avro version
-libraryDependencies += "org.apache.avro" % "avro-compiler" % "1.9.2"
```
brharrington, line 15: Nice simplification.

PR #250, opened 2020-08-10
atlas-persistence: unflatten datapoint list

jfz:
Currently data points from request is being add to queue one by one, this creates high sync contention for underlying BlockingQueue when data points per second reaches over 100K/s. With this change, only one data point list per request is passing through the queue, so it reduces sync contention significantly.
Also fixes some Scala warnings.

### DIFF
--- a/atlas-persistence/src/main/scala/com/netflix/atlas/persistence/LocalFilePersistService.scala
+++ b/atlas-persistence/src/main/scala/com/netflix/atlas/persistence/LocalFilePersistService.scala
@@ -38,6 +38,7 @@ import javax.inject.Inject
 import javax.inject.Singleton
 
 import scala.concurrent.Await
+import scala.concurrent.ExecutionContextExecutor
 import scala.concurrent.Future
 import scala.concurrent.duration.Duration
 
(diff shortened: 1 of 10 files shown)


atlas-persistence/src/main/scala/com/netflix/atlas/persistence/LocalFilePersistService.scala
```diff
@@ -68,21 +69,21 @@ class LocalFilePersistService @Inject()(
     fileConfig.getInt("avro-syncInterval")
   )
 
-  private var queue: SourceQueue[Datapoint] = _
+  private var queue: SourceQueue[List[Datapoint]] = _
```
jfz, line 72: This is the main change

project/BuildSettings.scala
```diff
@@ -6,7 +6,6 @@ object BuildSettings {
   val compilerFlags = Seq(
     "-deprecation",
     "-unchecked",
-    "-Xexperimental",
```
jfz, line 9: warning: -Xexperimental is deprecated: Not used since 2.13

PR #251, opened 2020-08-11
atlas-persistence: reject requests after queue closed

jfz:
During server or app shutdown, taking requests after queue closed causes data loss. Adding a check at Web API to reject requests after queue closed should fix this issue.

### DIFF
--- a/atlas-persistence/src/main/resources/application.conf
+++ b/atlas-persistence/src/main/resources/application.conf
@@ -8,7 +8,7 @@ atlas {
   }
 
   persistence {
-    queue-size = 100000
+    queue-size = 2000
     // Number of parallel file writing workers
     write-worker-size = 1
     local-file {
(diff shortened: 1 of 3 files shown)


atlas-persistence/src/main/resources/application.conf
```diff
@@ -8,7 +8,7 @@ atlas {
   }
 
   persistence {
-    queue-size = 100000
+    queue-size = 2000
```
jfz, line 11: Side change, default queue size can be smaller after un-flatten data list.

atlas-persistence/src/main/scala/com/netflix/atlas/persistence/PersistenceApi.scala
```diff
@@ -44,8 +44,12 @@ class PersistenceApi(localFileService: LocalFilePersistService) extends WebApi {
     parseEntity(customJson(p => PublishApi.decodeBatch(p))) {
       case Nil => complete(DiagnosticMessage.error(StatusCodes.BadRequest, "empty payload"))
       case dps: List[Datapoint] =>
-        localFileService.persist(dps)
-        complete(HttpResponse(StatusCodes.OK))
+        if (localFileService.isHealthy) {
+          localFileService.persist(dps)
+          complete(HttpResponse(StatusCodes.OK))
+        } else {
+          complete(DiagnosticMessage.error(StatusCodes.ServiceUnavailable, "service unhealthy"))
```
jfz, line 51: Client can retry in this case

brharrington:
Note, build is failing due to formatting issues.

PR #254, opened 2020-08-26
persistence: bucket files by start time to make range search easier

jfz:
Put files with the same start/end minute into a sub folder in s3. 
This makes file matching by range easier for downstream data processing pipeline. In theory it can create up to 1830(60+59+...+1) folders for an hour, but in reality I'd expect actual number to be much smaller because data points are mostly in order and range for a file is small, e.g. assume max range per file is 3 min, max num of folders <= 3*60.

### DIFF
--- a/atlas-persistence/src/main/scala/com/netflix/atlas/persistence/S3CopySink.scala
+++ b/atlas-persistence/src/main/scala/com/netflix/atlas/persistence/S3CopySink.scala
@@ -187,12 +187,13 @@ class S3CopySink(
         logger.info(s"copyToS3 done: file=$file, key=$s3Key")
       }
 
-      // Example file name: 2020-05-10T0300.i-localhost.1.XkvU3A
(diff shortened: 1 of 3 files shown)


atlas-persistence/src/main/scala/com/netflix/atlas/persistence/S3CopySink.scala
```diff
@@ -206,3 +207,19 @@ class S3CopySink(
     }
   }
 }
+
+object S3CopySink {
+
+  // Extract start-end minute from file name: "xyz.1200-1320" => "20-21"
+  // Use the special range "61-61" for tmp files.
```
brharrington, line 214: It would be helpful to add a note about what the `1200-1320` in the file name actually means. Also, based on the code it just divides by 60, so it would be `20-22`.

Can you use scaladoc style comment?

jfz: fixed

PR #255, opened 2020-09-03
druid: Add filter clause for non-zero metric value

bsyk:
Adding a filter on the metric value itself has a huge effect on query performance when there are sparce metrics.
Sparse metrics: Where a row has only 1 metric value defined, the others are null (or 0)
Sparse metrics combined with dimensions used across metrics, such as percentile, mean that we aggregate a lot of zeros, as well as grouping-by dimension values that may not be present in the final resultset.
Omitting these rows early with a filter significantly helps query performance in these situations.

### DIFF
--- a/atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
+++ b/atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
@@ -457,6 +457,11 @@ object DruidDatabaseActor {
         } else {
           val dimensions = getDimensions(simpleQuery, expr.finalGrouping)
 
+          // Add to the filter to remove rows that don't include the metric we're aggregating.
+          // This reduces what needs to be merged and passed back to the broker, improving query performance.
+          val metricValueFilter = Query.Not(Query.Equal(name, "0"))
+          val finalQueryWithFilter = finalQuery.and(metricValueFilter)
+
           // Filter out datapoints that are zero on the server side to reduce the payload
           // sizes and improve query performance.
           val havingSpec = HavingSpec("value", 0.0)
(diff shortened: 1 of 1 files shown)


atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
```diff
@@ -457,6 +457,11 @@ object DruidDatabaseActor {
...
           val dimensions = getDimensions(simpleQuery, expr.finalGrouping)
 
+          // Add to the filter to remove rows that don't include the metric we're aggregating.
+          // This reduces what needs to be merged and passed back to the broker, improving query performance.
+          val metricValueFilter = Query.Not(Query.Equal(name, "0"))
+          val finalQueryWithFilter = finalQuery.and(metricValueFilter)
+
           // Filter out datapoints that are zero on the server side to reduce the payload
           // sizes and improve query performance.
           val havingSpec = HavingSpec("value", 0.0)
```
samarthjain, line 467: I think we can get rid of the `havingSpec` which is applied post aggregation now that you have the same filter as `metricValueFilter` which is applied pre-aggregation.

atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
```diff
@@ -457,6 +457,11 @@ object DruidDatabaseActor {
         } else {
           val dimensions = getDimensions(simpleQuery, expr.finalGrouping)
 
+          // Add to the filter to remove rows that don't include the metric we're aggregating.
+          // This reduces what needs to be merged and passed back to the broker, improving query performance.
+          val metricValueFilter = Query.Not(Query.Equal(name, "0"))
```
brharrington, line 462: Note to self, Druid will handle the type coercion for the numeric fields:

https://druid.apache.org/docs/latest/querying/filters.html#filtering-on-numeric-columns

brharrington (approved):

atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
brharrington: Done.

PR #261, opened 2020-10-06
druid: Remove unneeded regex query when searching

bsyk:
The addition of the regex query adds ~25% to the query time and is not needed to perform the query that we're trying to do (get all possible tag values)

### DIFF
--- a/atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidClient.scala
+++ b/atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidClient.scala
@@ -263,7 +263,6 @@ object DruidClient {
     limit: Int = 1000
   ) {
     val queryType: String = "search"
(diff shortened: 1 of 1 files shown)


atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidClient.scala
```diff
@@ -263,7 +263,6 @@ object DruidClient {
     limit: Int = 1000
   ) {
     val queryType: String = "search"
-    val query: DruidQuery.type = DruidQuery
```
brharrington, line 266: This is here because the docs list `query` as being a required field for search queries.

https://druid.apache.org/docs/latest/querying/searchquery.html

However, it does seem to work fine without it so I guess the docs are wrong.

brharrington (approved):

PR #262, opened 2020-10-06
druid: Avoid using javascript filter

bsyk:
Use regex filter with lowercase extraction function and lowercase the regex query. Extraction functions are applied to the value prior to applying the filter, but the matched value is returned unchanged.

### DIFF
--- a/atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidFilter.scala
+++ b/atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidFilter.scala
@@ -44,7 +44,7 @@ object DruidFilter {
       case Query.LessThanEqual(k, v)    => js(k, "<=", v)
       case Query.HasKey(k)              => Not(Equal(k, "")) // druid: empty string is same as null
       case Query.Regex(k, v)            => Regex(k, s"^$v")
-      case Query.RegexIgnoreCase(k, v)  => reic(k, v)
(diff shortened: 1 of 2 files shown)


atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidFilter.scala
```diff
@@ -44,7 +44,7 @@ object DruidFilter {
       case Query.LessThanEqual(k, v)    => js(k, "<=", v)
       case Query.HasKey(k)              => Not(Equal(k, "")) // druid: empty string is same as null
       case Query.Regex(k, v)            => Regex(k, s"^$v")
-      case Query.RegexIgnoreCase(k, v)  => reic(k, v)
+      case Query.RegexIgnoreCase(k, v)  => Reic(k, s"^$v".toLowerCase())
```
brharrington, line 47: Performing `toLowerCase` on an arbitrary regex pattern is not guaranteed to be correct. For example inverse character classes (`\D` matches non-digits where `\d` matches digits) are often uppercase.

brharrington (changes requested):

atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidFilter.scala
bsyk: Ah yes, very good point. Don't know what I was thinking!
Took some digging, but I found a way to do case-insensitive searches directly.

brharrington (approved):

PR #263, opened 2020-10-06
druid: Filter null metrics

bsyk:
When we introduce new metric names, the value of that metric in older segments is null. When doing a group-by query, these segments without the metric will still perform aggregations. Filtering out nulls means the historical nodes can quickly remove rows that won't contribute


bsyk:
@vshirikian FYI

brharrington (approved):

PR #276, opened 2020-11-19
iep-lwc-fwding-admin: ConfigBinV2 migration

jfz:
migrate iep-lwc-fwding-admin to ConfigBinV2


iep-lwc-fwding-admin/src/main/resources/application.conf
```diff
@@ -29,7 +29,7 @@ iep.lwc.fwding-admin {
   edda-cache-refresh-interval = 30m
   fwd-metric-info-purge-limit = 5d
   cw-expr-uri = "http://localhost:7102/api/v2/cloudwatch-forwarding/clusters/%s"
-
+  user = user123
```
brharrington, line 32: Should this maybe default to something like the app name?

brharrington (approved):

PR #297, opened 2021-03-16
update to use AWS SDK v2

brharrington:
Most of the changes are just switching to the API
of of the V2 SDK. For DynamoDB uses, the changes
are a bit more involved because V2 doesn't support
the document level API that was being used.


jfz (approved):

copperlight (approved):

copperlight:
That was a fair bit of work, porting the Slotting Service over to V2. Thanks for getting that done!

PR #319, opened 2021-07-29
[patch] Use limit only if less than Int.MAX_VALUE

bsyk:
This is a requirement for Druid TopN queries, they are rejected if >= Int.MAX_VALUE


brharrington (approved):
This isn't strictly correct for the query semantics of foreach, but is probably the best option for now.

PR #364, opened 2022-06-17
druid: add support for histogram type

brharrington:
Updates druid integration to support the custom spectator
histogram type. This type encodes the percentile distribution
data as the value in druid rather than relying on the normal
approach of having a dimension for the bucket index and a
simple double value. The segment metadata will indicate if it
is a timer or distribution summary and the `spectatorHistogram`
aggregation can be used to merge entries on the druid side.
The result will get mapped by the client to the previous format
so that all processing on top of it remains the same.

The histogram type also works with simple aggregates like
`doubleSum` so that use-cases for determining the overall
updates per second based on summing the individual buckets will
still work as expected. However, queries that explicitly
restrict based on the `percentile` dimension will not work.
That is not a common use-case and is not expected to be an
issue.


atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidClient.scala
```diff
@@ -201,26 +211,58 @@ class DruidClient(
...
+
   private def isNullOrEmpty(v: String): Boolean = v == null || v.isEmpty
 }
 
 object DruidClient {
 
   case class Datasource(dimensions: List[String], metrics: List[Metric])
 
-  case class Metric(name: String, dataType: String) {
+  case class Metric(name: String, dataType: String = "DOUBLE") {
```
manolama, line 252: Enum for the data type might be a teensy bit more efficient.

manolama (approved):

PR #365, opened 2022-07-18
add memorydb metrics

copperlight:
This implementation is similar the the elasticache redis metrics, with
metrics selection and naming following the existing conventions.
    
This covers the most useful host-level and cache metrics, and all of
the commandstats metrics.

### DIFF
--- a/atlas-cloudwatch/src/main/resources/reference.conf
+++ b/atlas-cloudwatch/src/main/resources/reference.conf
@@ -369,6 +369,7 @@ atlas {
       "kinesis",
       "lambda",
       "lambda-fn-res",
+      "memorydb",
       "nat-gateway",
       "neptune-instance",
       "neptune-cluster-role",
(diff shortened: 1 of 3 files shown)


atlas-cloudwatch/src/main/resources/reference.conf
```diff
@@ -411,6 +411,7 @@ include "events.conf"
 include "iot.conf"
 include "kinesis.conf"
 include "lambda.conf"
+include "memorydb.conf"
```
brharrington, line 414: Should it also be included in the `categories` list above?

brharrington (approved):

atlas-cloudwatch/src/main/resources/reference.conf
copperlight: Yes, good catch.

PR #366, opened 2022-07-19
add aws.memorydb tag

copperlight:
This maps the ClusterName dimension to the aws.memorydb tag. The NodeId
dimension is already mapped to the nf.node tag.

Considering this mapping, the elasticache aws.node dimension is better
represented by nf.node, which is more in-line with existing common tags.

Add the EngineCPUUtilization metric, which is recommended for use when
there are four or more vCPU.

### DIFF
--- a/atlas-cloudwatch/src/main/resources/reference.conf
+++ b/atlas-cloudwatch/src/main/resources/reference.conf
@@ -164,7 +164,7 @@ atlas {
         },
         {
           name = "CacheNodeId"
-          alias = "aws.node"
+          alias = "nf.node"
         },
         {
           name = "ClientId"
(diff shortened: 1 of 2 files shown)


atlas-cloudwatch/src/main/resources/reference.conf
```diff
@@ -164,7 +164,7 @@ atlas {
         },
         {
           name = "CacheNodeId"
-          alias = "aws.node"
+          alias = "nf.node"
```
brharrington, line 167: This is changing existing behavior so may cause some existing usages to break.

copperlight: Yes, I had considered that. I was thinking we could fix-forward, since this ends up being a more consistent configuration.

A quick check of atlas-dashboard and sourcegraph did not turn up any uses of this particular tag.

There are a fair number of elasticache clusters defined at this point, but there are only two uses of the `aws.node` tag in alert-configs and no uses in dashboards. We can fix these when this change rolls out.

PR #367, opened 2022-08-05
lwc-bridge: prefer tags on metric to common tags

brharrington:
If there is a conflict between the common tags and the
tags on a metric for a batch payload, then prefer the
tags local to the metric. This matches the precedence
behavior for the backends.


manolama (approved):
LGTM

PR #374, opened 2022-09-09
druid: fix UOE when using not clauses

brharrington:
Fixes UnsupportedOperationException when using NOT clauses with either the `nf.datasource` or `name` dimensions.


manolama (commented):
LGTM

PR #377, opened 2022-09-23
add isDisabled flag to asg details

copperlight:
This flag reports whether or not the ASG is disabled, according to Netflix operational standards. The Launch process must be suspended on the ASG for this to be true.

Since we are adding one field to the ASG details, there will be a difference between data stored in DynamoDB and newly crawled data. This will trigger a merge on the first startup of the new code, which will preserve slot numbers for all existing instances. Future starts of the new code will update timestamps for unchanged ASGs.

### DIFF
--- a/atlas-slotting/src/main/scala/com/netflix/atlas/slotting/Grouping.scala
+++ b/atlas-slotting/src/main/scala/com/netflix/atlas/slotting/Grouping.scala
@@ -15,8 +15,6 @@
  */
 package com.netflix.atlas.slotting
 
-import java.nio.ByteBuffer
-import java.time.Instant
 import com.netflix.atlas.json.Json
 import com.netflix.spectator.api.Counter
 import com.netflix.spectator.ipc.ServerGroup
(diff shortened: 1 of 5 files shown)


atlas-slotting/src/main/scala/com/netflix/atlas/slotting/Grouping.scala
```diff
@@ -145,6 +149,22 @@ trait Grouping extends StrictLogging {
...
+    *     An AWS AutoScalingGroup.
+    * @return
+    *     The disabled state of the ASG.
+    */
+  def mkIsDisabled(asg: AutoScalingGroup): Boolean = {
+    asg
+      .suspendedProcesses()
+      .asScala
+      .exists { p =>
+        p.getValueForField("ProcessName", classOf[String]).toScala.contains("Launch")
```
brharrington, line 164: Why not just use the field accessor?

```scala
.exists(_.processName == "Launch")
```

copperlight: Good point, much cleaner this way.

PR #389, opened 2023-02-21
fix aws direct connect metrics

copperlight:
The affected metrics are all reported as aggregates (averages) over a five minute period by default.


atlas-cloudwatch/src/main/resources/dx.conf
```diff
@@ -27,7 +27,7 @@ atlas {
         {
           name = "ConnectionBpsEgress"
           alias = "aws.dx.connectionBytes"
-          conversion = "sum",
+          conversion = "sum,rate",
```
brharrington, line 30: Note, if the unit in CW is a rate as indicated by the docs, then this will be a no-op.

brharrington (approved):

PR #401, opened 2023-03-07
Atlas Cloudwatch: Add the FirehoseMetric case class for storing info

manolama:
parsed from the firehose Cloud Watch stream.
Also add an OpenHFT XXHash implementation utility for quickly calculating a unique hash on the Cloud Watch metric for caching.

### DIFF
--- a/atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/FirehoseMetric.scala
+++ b/atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/FirehoseMetric.scala
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2014-2023 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.atlas.cloudwatch
+
+import com.netflix.atlas.util.XXHasher
+import software.amazon.awssdk.services.cloudwatch.model.Datapoint
+import software.amazon.awssdk.services.cloudwatch.model.Dimension
+
+/**
+  * Container to hold a metric parsed from a Firehose Cloud Watch metric.
+  */
+case class FirehoseMetric(
+  metricStreamName: String,
+  namespace: String,
+  metricName: String,
+  dimensions: List[Dimension],
+  datapoint: Datapoint
+) {
+
+  def xxHash: Long = {
+    var hash = XXHasher.hash(namespace)
+    hash = XXHasher.updateHash(hash, metricName)
+    dimensions.sortBy(_.name()).foreach { d =>
+      hash = XXHasher.updateHash(hash, d.name())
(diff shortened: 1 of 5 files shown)


atlas-cloudwatch/src/main/scala/com/netflix/atlas/util/XXHasher.scala
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+package com.netflix.atlas.util
+
+/**
+  * Simple wrapper and utility functions for hashing with the OpenHFT XXHash implementation.
+  */
+object XXHasher {
+
+  private val hasher = LongHashFunction.xx
```
brharrington, line 23: I think the build failure may be that `LongHashFunction` wasn't imported.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/FirehoseMetric.scala
```diff
@@ -0,0 +1,42 @@
...
+  namespace: String,
+  metricName: String,
+  dimensions: List[Dimension],
+  datapoint: Datapoint
+) {
+
+  def xxHash: Long = {
+    var hash = XXHasher.hash(namespace)
+    hash = XXHasher.updateHash(hash, metricName)
+    dimensions.sortBy(_.name()).foreach { d =>
```
brharrington, line 36: How often do you expect this hash will be computed for a given object? It might be better to pre-sort the list when creating the object.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/FirehoseMetric.scala
```diff
@@ -0,0 +1,42 @@
...
+  */
+case class FirehoseMetric(
+  metricStreamName: String,
+  namespace: String,
+  metricName: String,
+  dimensions: List[Dimension],
+  datapoint: Datapoint
+) {
+
+  def xxHash: Long = {
```
brharrington, line 33: Why is this needed for this use-case as opposed to the simple built in hash? Maybe add that context as a comment.

atlas-cloudwatch/src/test/scala/com/netflix/atlas/util/XXHasherSuite.scala
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ */
+package com.netflix.atlas.util
+
+import munit.FunSuite
+
+class XXHasherSuite extends FunSuite {
+
+  test("hash byte array") {
+    assertEquals(XXHasher.hash("Hello World".getBytes("UTF-8")), 7148569436472236994L)
```
brharrington, line 23: Is the xx hash implementation guaranteed to be stable? If they tweak values over time to improve characteristics these tests could be flakey for any updates.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/util/XXHasher.scala
manolama: Yup, copy pasta before adding the dep.

atlas-cloudwatch/src/test/scala/com/netflix/atlas/util/XXHasherSuite.scala
manolama: Yes, it will be stable and if something _does_ inadvertently change then we want these UTs to fail so we can roll back or choose a different implementation going forward.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/FirehoseMetric.scala
manolama: Done.

brharrington (approved):

PR #402, opened 2023-03-07
Atlas Cloud Watch: Add the PublishQueue class to replace the Client A…

manolama:
…ctor

grouping and flushing data points to Publish Proxy instances. A queue will be instantiated for each stack and a router (in the next PR) wil use the AWS account to route data points to the proper queue.

### DIFF
--- a/atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/PublishQueue.scala
+++ b/atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/PublishQueue.scala
@@ -1,0 +1,183 @@
+/*
+ * Copyright 2014-2023 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.atlas.cloudwatch
+
+import akka.actor.ActorSystem
+import akka.http.scaladsl.model.HttpEntity
+import akka.http.scaladsl.model.HttpMethods
+import akka.http.scaladsl.model.HttpRequest
+import akka.http.scaladsl.model.HttpResponse
+import akka.stream.scaladsl.Keep
+import akka.stream.scaladsl.Sink
+import akka.util.ByteString
+import com.netflix.atlas.akka.AkkaHttpClient
+import com.netflix.atlas.akka.CustomMediaTypes
+import com.netflix.atlas.akka.StreamOps
+import com.netflix.atlas.json.Json
+import com.netflix.atlas.poller.Messages
+import com.netflix.atlas.poller.Messages.MetricsPayload
+import com.netflix.spectator.api.Id
+import com.netflix.spectator.api.Registry
+import com.typesafe.config.Config
+import com.typesafe.scalalogging.StrictLogging
+
+import java.time.Duration
+import java.util.concurrent.ScheduledExecutorService
+import java.util.concurrent.TimeUnit
+import scala.concurrent.duration.FiniteDuration
+import scala.util.Failure
+import scala.util.Success
+
+/**
+  * Simple queue for batching data points to be sent to publish proxy. Retries occur only on 429s, 504s or exceptions
+  * from the client.
+  */
+class PublishQueue(
+  config: Config,
+  registry: Registry,
+  val stack: String,
+  val uri: String,
+  httpClient: AkkaHttpClient,
+  scheduler: ScheduledExecutorService
+)(implicit system: ActorSystem)
+    extends StrictLogging {
+
+  private implicit val executionContext = system.dispatcher
+
+  private val datapointsDropped =
+    registry.createId("atlas.cloudwatch.queue.dps.dropped", "stack", stack)
+  private val datapointsSent = registry.counter("atlas.cloudwatch.queue.dps.sent", "stack", stack)
+  private val retryAttempts = registry.counter("atlas.cloudwatch.queue.retries", "stack", stack)
+  private val droppedFull = registry.counter(datapointsDropped.withTags("reason", "queueFull"))
+  private val droppedRetries = registry.counter(datapointsDropped.withTags("reason", "maxRetries"))
+
+  private val maxRetries = getSetting("maxRetries")
+  private val queueSize = getSetting("queueSize")
+  private val batchSize = getSetting("batchSize")
+  private val batchTimeout = getDurationSetting("batchTimeout")
+
+  private[cloudwatch] val publishQueue = StreamOps
+    .blockingQueue[AtlasDatapoint](registry, s"${stack}PubQueue", queueSize)
+    .groupedWithin(batchSize, FiniteDuration.apply(batchTimeout.getSeconds, TimeUnit.SECONDS))
+    .map(publish)
+    .toMat(Sink.ignore)(Keep.left)
+    .run()
+
+  /**
+    * Adds the data point to the queue if there is room. Increments a metric if not.
+    *
+    * @param datapoint
+    *     The non-null data point to enqueue.
+    */
+  def enqueue(datapoint: AtlasDatapoint): Unit = {
+    if (!publishQueue.offer(datapoint)) {
+      droppedFull.increment()
+    }
+  }
+
+  private[cloudwatch] def publish(datapoints: Seq[AtlasDatapoint]): Unit = {
+    val size = datapoints.size
+    val payload = Json.smileEncode(MetricsPayload(Map.empty, datapoints))
+    datapointsSent.increment(size)
+    publish(payload, 0, size)
+  }
+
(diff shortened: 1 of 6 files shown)


atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/PublishQueue.scala
```diff
@@ -0,0 +1,183 @@
...
+
+  /**
+    * Adds the data point to the queue if there is room. Increments a metric if not.
+    *
+    * @param datapoint
+    *     The non-null data point to enqueue.
+    */
+  def enqueue(datapoint: AtlasDatapoint): Unit = {
+    if (!publishQueue.offer(datapoint)) {
+      droppedFull.increment()
```
brharrington, line 87: There are standard metrics on the queue for reporting number of items submitted or dropped due to being full.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/PublishQueue.scala
```diff
@@ -0,0 +1,183 @@
...
+        response.status.intValue() match {
+          case 200 => // All is well
+            response.discardEntityBytes()
+          case 202 | 206 => // Partial failure
+            val id = datapointsDropped.withTag("reason", "partialFailure")
+            incrementFailureCount(id, response, size)
+          case 400 => // Bad message, all data dropped
+            val id = datapointsDropped.withTag("reason", "completeFailure")
+            incrementFailureCount(id, response, size)
+          case 429 | 504 => // backoff
```
brharrington, line 118: Was this supposed to be 503? We don't typically retry on timeouts because you cannot guarantee some of the data wasn't processed.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/PublishQueue.scala
```diff
@@ -0,0 +1,183 @@
...
+            retry(payload, retries, size)
```
brharrington, line 119: This could be done using [max-retries](https://doc.akka.io/docs/akka-http/current/client-side/host-level.html#retrying-a-request) setting on the client rather than explicitly here.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/PublishQueue.scala
```diff
@@ -0,0 +1,183 @@
...
+  private val droppedRetries = registry.counter(datapointsDropped.withTags("reason", "maxRetries"))
+
+  private val maxRetries = getSetting("maxRetries")
+  private val queueSize = getSetting("queueSize")
+  private val batchSize = getSetting("batchSize")
+  private val batchTimeout = getDurationSetting("batchTimeout")
+
+  private[cloudwatch] val publishQueue = StreamOps
+    .blockingQueue[AtlasDatapoint](registry, s"${stack}PubQueue", queueSize)
+    .groupedWithin(batchSize, FiniteDuration.apply(batchTimeout.getSeconds, TimeUnit.SECONDS))
```
brharrington, line 74: I would recommend doing:

```
FiniteDuration(batchTimeout.toNanos, TimeUnit.NANOSECONDS)
```

This avoids surprises if say we used a duration of 500ms.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/PublishQueue.scala
```diff
@@ -0,0 +1,183 @@
...
+            retry(payload, retries, size)
+          case v => // Unexpected, assume all dropped
+            response.discardEntityBytes()
+            val id = datapointsDropped.withTag("reason", s"status_$v")
+            registry.counter(id).increment(size)
+        }
+
+      case Failure(ex) =>
+        logger.error(s"Failed publishing to ${uri} for ${stack}", ex)
+        incrementException(ex)
```
brharrington, line 128: This is already covered by common IPC client metrics.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/poller/Messages.scala
```diff
@@ -42,7 +42,10 @@ object Messages {
     * @param metrics
     *     Metrics collected by the poller.
     */
-  case class MetricsPayload(tags: Map[String, String] = Map.empty, metrics: List[Datapoint] = Nil)
+  case class MetricsPayload(
+    tags: Map[String, String] = Map.empty,
+    metrics: Iterable[Datapoint] = Nil
```
brharrington, line 47: Why was this changed to `Iterable`?

atlas-cloudwatch/src/test/scala/com/netflix/atlas/cloudwatch/PublishQueueSuite.scala
```diff
@@ -0,0 +1,301 @@
...
+    val queue =
+      new PublishQueue(config, registry, "main", "http://localhost", httpClient, scheduler)
+    queue.publish(
+      Seq(
+        Datapoint(Map("k1" -> "v1"), timestamp, 42.5),
+        Datapoint(Map("k2" -> "v2"), timestamp, 24.1)
+      )
+    )
+
+    Thread.sleep(threadSleep)
```
brharrington, line 82: Sleeps tend to make tests really flakey and should be avoided.

atlas-cloudwatch/src/test/scala/com/netflix/atlas/cloudwatch/PublishQueueSuite.scala
```diff
@@ -0,0 +1,301 @@
...
+    droppedPartial: Long = 0,
+    droppedComplete: Long = 0,
+    dropped500: Long = 0,
+    droppedQueueFull: Long = 0,
+    droppedRetries: Long = 0,
+    exceptions: Long = 0
+  ): Unit = {
+    assertEquals(registry.counter("atlas.cloudwatch.queue.dps.sent", "stack", "main").count, sent)
+    assertEquals(registry.counter("atlas.cloudwatch.queue.retries", "stack", "main").count, retries)
+    assertEquals(
```
brharrington, line 230: This could be refactored to be less repetitive, maybe something like:

```scala
def assertCounters(
    sent: Long = 0,
    retries: Long = 0,
    exceptions: Long = 0
    dropped: Map[String, Long]
): Unit = {
  // check others
  List("partialFaliure", "completeFailure", "status_500", "queueFull", "maxRetries").foreach { reason =>
    assertEquals(
      registry
        .counter("atlas.cloudwatch.queue.dps.dropped", "stack", "main", "reason", reason)
        .count,
      dropped.getOrElse(reason, 0L)
    )
  }
}
```

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/PublishQueue.scala
manolama: It was a 504 in the old pub proxy. I can drop it.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/PublishQueue.scala
manolama: The docs say Akka won't retry a `POST` though. And I do like having the counter to see if anything is retrying. Is it OK to keep this with just the backoff?

atlas-cloudwatch/src/main/scala/com/netflix/atlas/poller/Messages.scala
manolama: The queue grouping returns a `Seq` instead of a `List`.

atlas-cloudwatch/src/test/scala/com/netflix/atlas/cloudwatch/PublishQueueSuite.scala
manolama: Hmm, nice.

atlas-cloudwatch/src/test/scala/com/netflix/atlas/cloudwatch/PublishQueueSuite.scala
manolama: Agreed, I hate it but still haven't dug into the Akka framework to figure out how to tick through the dispatcher to make sure the futures are executed.
I could refactor it for now to return a future privately.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/PublishQueue.scala
manolama: Ack, I can drop it. I like having the `stack` tag to correlate with the dropped metrics.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/PublishQueue.scala
```diff
@@ -104,33 +102,44 @@ class PublishQueue(
...
           case 400 => // Bad message, all data dropped
             val id = datapointsDropped.withTag("reason", "completeFailure")
-            incrementFailureCount(id, response, size)
-          case 429 | 504 => // backoff
+            incrementFailureCount(id, response, size, promise)
+          case 429 => // backoff
             retry(payload, retries, size)
-          case v => // Unexpected, assume all dropped
             response.discardEntityBytes()
+            promise.complete(Try(null))
```
brharrington, line 121: There is a [NotUsed](https://doc.akka.io/api/akka/current/akka/NotUsed.html) type in Akka that is typically use for this sort of response where the data value is not used.

brharrington (approved):

PR #414, opened 2023-03-10
Atlas Cloud Watch: Add the Redis cluster metrics processor that stores

manolama:
and retrieves data from a Redis cluster.

### DIFF
--- a/atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/RedisClusterCloudWatchMetricsProcessor.scala
+++ b/atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/RedisClusterCloudWatchMetricsProcessor.scala
@@ -1,0 +1,386 @@
+/*
+ * Copyright 2014-2023 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.atlas.cloudwatch
+
+import akka.actor.ActorSystem
+import com.netflix.atlas.cloudwatch.CloudWatchMetricsProcessor.expirationSeconds
+import com.netflix.atlas.cloudwatch.CloudWatchMetricsProcessor.newCacheEntry
+import com.netflix.atlas.cloudwatch.RedisClusterCloudWatchMetricsProcessor.getKey
+import com.netflix.iep.leader.api.LeaderStatus
+import com.netflix.spectator.api.Registry
+import com.typesafe.config.Config
+import com.typesafe.scalalogging.StrictLogging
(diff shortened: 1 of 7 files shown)


atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/RedisClusterCloudWatchMetricsProcessor.scala
```diff
@@ -0,0 +1,386 @@
...
+import scala.concurrent.Promise
+import scala.util.Failure
+import scala.util.Success
+import scala.util.Try
+import scala.util.Using
+
+/**
+  * A Redis cluster implementation of the Cloud Watch Metrics processor.
+  *
+  * Keys are distributed across the cluster based on the XX hash of the AWS data. Scraping works by finiding the leaders
```
brharrington, line 51: `finding`

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/RedisClusterCloudWatchMetricsProcessor.scala
```diff
@@ -0,0 +1,386 @@
...
+                    batches += getAndPublish(node, keys, ts)
+                  })
+
+                  Future.sequence(batches).onComplete {
+                    case Success(_) =>
+                      batchPromise.complete(Try(null))
+                    case Failure(ex) =>
+                      // shouldn't happen.
+                      logger.error("Failed on one or more batches", ex)
+                      registry.counter(batchFailure.withTag("node", node)).increment()
```
brharrington, line 197: Generally it is not a good idea to have remote node dimension on metrics as the cross product with the local node can be a big multiple. That level of detail is probably a better fit for the log message.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/RedisClusterCloudWatchMetricsProcessor.scala
```diff
@@ -0,0 +1,386 @@
...
+    try {
+      val hash = datapoint.xxHash
+      val key = getKey(hash)
+      val existing =
+        try {
+          jedis.get(key)
+        } catch {
+          case ex: Exception =>
+            // TODO - watch out, this logging could be really nasty if the cluster or node is down
+            logger.debug("Failed to get key from Redis", ex)
```
brharrington, line 115: Is the case where logging more useful a particular exception type? It might be helpful to know the datapoint/key in the log if debugging.

brharrington (approved):

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/RedisClusterCloudWatchMetricsProcessor.scala
manolama: Possibly, though it's more likely that fetches for all keys in a given slot or on a node will fail than an individual key. I should add the slot and key at least.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/RedisClusterCloudWatchMetricsProcessor.scala
manolama: Aye. I was using this to validate key distribution across nodes. The cardinality here should be pretty small though at any given time, just 1 scraping node by the # of leader Redis nodes, so 2 to 5 initially. Per region. Mind if I keep it for now, validate the distribution, then remove it?

PR #415, opened 2023-03-10
Atlas Cloud Watch: Add the Firehose endpoint that will receive messag…

manolama:
…es from

Firehose, parse and process them. Includes the RequestId class that stores the timestamp and request ID that firehose requires in the response to accept the data as successfully received.

### DIFF
--- a/atlas-cloudwatch/src/main/scala/com/netflix/atlas/webapi/CloudWatchFirehoseEndpoint.scala
+++ b/atlas-cloudwatch/src/main/scala/com/netflix/atlas/webapi/CloudWatchFirehoseEndpoint.scala
@@ -1,0 +1,267 @@
+/*
+ * Copyright 2014-2023 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


atlas-cloudwatch/src/main/scala/com/netflix/atlas/webapi/CloudWatchFirehoseEndpoint.scala
```diff
@@ -0,0 +1,267 @@
...
+  private val parseException = registry.createId("atlas.cloudwatch.firehose.parse.exception")
+
+  override def routes: Route = {
+    post {
+      endpointPath("api" / "v1" / "firehose") {
+        handleReq
+      } ~
+      complete(StatusCodes.NotFound)
+    } ~
+    put { // just to be safe
```
brharrington, line 63: I think you could avoid the redundancy with something like:

```scala
(post | put) {
  // ...
}
```

brharrington (approved):

atlas-cloudwatch/src/main/scala/com/netflix/atlas/webapi/CloudWatchFirehoseEndpoint.scala
manolama: oh nice.

PR #426, opened 2023-03-22
Atlas Cloudwatch: Backport the `MetricCategory.dimensionsMatch` funct…

manolama:
…ion.

Then modify the ListMetricsActor to filter on an exact match with the dimensions listed in the category.


atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/MetricCategory.scala
```diff
@@ -101,6 +102,25 @@ case class MetricCategory(
...
+    * True if the tags match those of the category config, false if there were fewer or
+    * more tags than expected.
+    */
+  def dimensionsMatch(tags: List[Dimension]): Boolean = {
+    var matched = 0
+    var extraTag = false
+    tags.foreach { d =>
+      if (dimensions.contains(d.name())) matched += 1 else extraTag = true
+    }
+    !extraTag && matched == dimensions.size
```
brharrington, line 122: This function could be rewritten as:

```scala
tags.forall(d => dimensions.contains(d.name))
```

brharrington (approved):

PR #447, opened 2023-04-08
atlas cloudwatch: More config fixes.

manolama:


atlas-cloudwatch/src/main/resources/firehose.conf
```diff
@@ -106,7 +96,7 @@ atlas {
         {
           name = "DeliveryToHttpEndpoint.DataFreshness"
           alias = "aws.firehose.data.freshNess"
```
brharrington, line 98: Why is the `N` capitalized?

manolama: Doh, fixing.

PR #454, opened 2023-05-09
Add support for querying HLLSketch types

bsyk:
These represent approximate distinct counts. We can treat them as longsum/counters. Usage will be nuanced, we'll probably need to ensure callers are performing per-step queries for the results to make sense.


atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
```diff
@@ -439,6 +439,7 @@ object DruidDatabaseActor {
       case Histogram(_) if metric.isDistSummary => Aggregation.distSummary(metric.name)
       case DataExpr.GroupBy(e, _)               => toAggregation(metric, e)
       case DataExpr.Consolidation(e, _)         => toAggregation(metric, e)
+      case _: DataExpr.Sum if metric.isSketch   => Aggregation.distinct(metric.name)
```
brharrington, line 442: So if other aggregates are used with this type, what will the behavior be?

bsyk: Yes, that's a good point. We wouldn't do anything different if called with max or min. I'll change to always use this aggregator for columns of this type.

brharrington (approved):

PR #460, opened 2023-06-01
druid: fix OR query using name in tags queries

bsyk:
Similar fix to #459

### DIFF
--- a/atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
+++ b/atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
@@ -45,6 +45,7 @@ import com.netflix.spectator.impl.PatternMatcher
 import com.typesafe.config.Config
 import com.typesafe.scalalogging.StrictLogging
 
+import java.time.temporal.ChronoUnit
 import scala.concurrent.duration._
 import scala.util.Failure
 import scala.util.Success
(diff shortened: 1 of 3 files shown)


atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
```diff
@@ -215,25 +216,44 @@ class DruidDatabaseActor(config: Config) extends Actor with StrictLogging {
...
-            .map(_.flatMap(_.values))
-            .fold(List.empty[String]) { (vs1, vs2) =>
-              ListHelper.merge(tq.limit, vs1, vs2)
+          val druidQueries = datasources
+            .flatMap { ds =>
+              // If there's a metric name in the query, ensure we simplify the query taking
+              // into account the name match earlier so only the branch of the :or for the particular
+              // metric will be present when constructing the filter
+              val metric = if (Query.allKeys(query).contains("name")) {
+                ds.metrics.filter(m => query.couldMatch(m.tags)).map { m => m.metric.name }.head
```
brharrington, line 225: I'm not sure this is correct as there could be an OR clause with multiple names from the same datasource.

brharrington: I think it would still be an issue because you could have additional clauses on each. For example, consider an expression like:

```
name,a,:eq,dim,1,:eq,:and,
name,b,:eq,dim,2,:eq,:and,:or
```

Assuming `a` and `b` are for the same datasource, lets say `a` gets chosen it would get simplified to:

```
:true,dim,1,:eq,:and,
:false,dim,2,:eq,:and,:or

=>

dim,1,:eq,
:false,:or

=>

dim,1,:eq
```

In this case, wouldn't the desired result for the tag query be `dim,1,:eq,dim,2,:eq,:or`?

bsyk: Yes, you're right. I'd only tested very simple cases.

brharrington (approved):

PR #464, opened 2023-06-08
atlas cloudwatch: update media metrics.

manolama:
Add metrics for mediaconnect and medialive.

### DIFF
--- a/atlas-cloudwatch/src/main/resources/medialive.conf
+++ b/atlas-cloudwatch/src/main/resources/medialive.conf
@@ -90,6 +90,11 @@ atlas {
           tags = []
         },
         {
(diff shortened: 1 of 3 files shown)


atlas-cloudwatch/src/main/resources/medialive.conf
```diff
@@ -89,6 +89,11 @@ atlas {
           conversion = "min"
           tags = []
         },
+        {
+          name = "FillMsec"
+          alias = "aws.medialive.fillMsec"
```
brharrington, line 94: The name here is likely misleading, if the unit is specified on the CW metric it will automatically get converted to a base unit of seconds.

Issue #462, opened 2023-06-01
seeing spurious test failures from atlas-cloudwatch


brharrington:
```
com.netflix.atlas.cloudwatch.AwsConfigAccountSupplierSuite:
==> X com.netflix.atlas.cloudwatch.AwsConfigAccountSupplierSuite.success  1.155s scala.reflect.internal.Symbols$CyclicReference: illegal cyclic reference involving class SelectAggregateResourceConfigResponse
==> X com.netflix.atlas.cloudwatch.AwsConfigAccountSupplierSuite.empty response  0.018s java.lang.AssertionError: assertion failed: no symbol could be loaded from interface software.amazon.awssdk.services.config.model.SelectAggregateResourceConfigResponse$Builder in class SelectAggregateResourceConfigResponse with name Builder and classloader sbt.internal.LayeredClassLoader@749f5836
    at scala.reflect.internal.SymbolTable.throwAssertionError(SymbolTable.scala:171)
    at scala.reflect.runtime.JavaMirrors$JavaMirror.classToScala1(JavaMirrors.scala:1083)
    at scala.reflect.runtime.JavaMirrors$JavaMirror.$anonfun$classToScala$1(JavaMirrors.scala:1041)
    at scala.reflect.runtime.JavaMirrors$JavaMirror.$anonfun$toScala$1(JavaMirrors.scala:152)
    at scala.reflect.runtime.TwoWayCaches$TwoWayCache.toScala(TwoWayCaches.scala:50)
    at scala.reflect.runtime.JavaMirrors$JavaMirror.toScala(JavaMirrors.scala:150)
    at scala.reflect.runtime.JavaMirrors$JavaMirror.classToScala(JavaMirrors.scala:1041)
    at scala.reflect.runtime.JavaMirrors$JavaMirror.typeToScala(JavaMirrors.scala:1149)
    at scala.reflect.runtime.JavaMirrors$JavaMirror.targToScala$1(JavaMirrors.scala:1136)
    at scala.reflect.runtime.JavaMirrors$JavaMirror.$anonfun$targsToScala$3(JavaMirrors.scala:1138)
    at scala.reflect.runtime.JavaMirrors$JavaMirror.targsToScala(JavaMirrors.scala:1138)
    at scala.reflect.runtime.JavaMirrors$JavaMirror.typeToScala(JavaMirrors.scala:1162)
[...]

PR #467, opened 2023-06-09
cloudwatch: attempt mockito fix for #462

manolama:
Mockito-scala docs excplicity ask to not declare a dependency on mockito-core, instead let mockito-scala bring it in. I was likely including a much older version than needed and could have caused load ordering issues.


brharrington:
Could we just get rid of mockito entirely?

manolama:
yeah but I want to test those blasted AWS calls. I'm poking around more and I see a local dynamo implementation for testing but I don't know about the others. I'll keep poking.

manolama:
Will try yanking entirely then.

PR #471, opened 2023-06-19
atlas cloudwatch: Add FSx for OnTap metrics.

manolama:


atlas-cloudwatch/src/main/resources/fsx.conf
```diff
@@ -0,0 +1,257 @@
...
+      period = 1m
+      end-period-offset = 1
+
+      dimensions = [
+        "FileSystemId"
+      ]
+
+      metrics = [
+        {
+          name = "DataReadBytes"
```
brharrington, line 17: It looks like this was mapped in below with dimensions `FileSystemId` and `VolumeId`. We typically only map it in once with the most specific set of dimensions rather than have duplicates.

atlas-cloudwatch/src/main/resources/fsx.conf
```diff
@@ -0,0 +1,257 @@
...
+
+      metrics = [
+        {
+          name = "StorageCapacity"
+          alias = "aws.fsx.detailedFilesystem.storageCapacity"
+          conversion = "max"
+        },
+        {
+          name = "StorageUsed"
+          alias = "aws.fsx.detailedFilesystem..storageUsed"
```
brharrington, line 103: `s/.././`

atlas-cloudwatch/src/main/resources/fsx.conf
```diff
@@ -0,0 +1,257 @@
...
+
+      dimensions = [
+        "FileSystemId",
+        "StorageTier",
+        "DataType"
+      ]
+
+      metrics = [
+        {
+          name = "StorageCapacity"
```
brharrington, line 97: These also seem to be duplicates of some below with more specific dimensions.

atlas-cloudwatch/src/main/resources/fsx.conf
```diff
@@ -0,0 +1,257 @@
...
+          alias = "aws.fsx.detailedVolume.storageCapacity"
+          conversion = "max"
+        },
+        {
+          name = "StorageUsed"
+          alias = "aws.fsx.detailedVolume..storageUsed"
```
brharrington, line 251: `s/.././`

atlas-cloudwatch/src/main/resources/fsx.conf
manolama: ack, I can drop the FS bits then.

Also, we're kinda stuck with `FileSystemId` mapping to `aws.efs` which it's not really.

atlas-cloudwatch/src/main/resources/fsx.conf
manolama: yeah, I'm waffling on the detailed vs "free" metrics. Checking the console we had detailed enabled.

PR #474, opened 2023-06-23
atlas cloudwatch: remove polling filter.

manolama:
Allows for polling more frequently, e.g. every 5m for the EC2 VPC metrics.


atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchPoller.scala
```diff
@@ -102,16 +102,13 @@ class CloudWatchPoller(
   private val dpsPolled = registry.counter("atlas.cloudwatch.poller.dps.polled")
   private val frequency = config.getDuration("atlas.cloudwatch.poller.frequency").getSeconds
 
-  private val periodFilter =
-    config.getDuration("atlas.cloudwatch.poller.period-filter").getSeconds.toInt
```
brharrington, line 106: Please remove the setting from the config as well to avoid confusion in the future:

https://github.com/Netflix-Skunkworks/iep-apps/blob/main/atlas-cloudwatch/src/main/resources/reference.conf#L122

brharrington:
Build issue should be fixed in #475 if you rebase.

PR #491, opened 2023-09-28
slotting: add IPv6 address to response

brharrington:
Add IPv6 address to the instance details. Also make the private IP address optional as it is null on instances running in an IPv6 only subnet.


copperlight (approved):
This is clean. It was more work updating the tests than the actual code!

PR #501, opened 2023-11-02
atlas cloudwatch: add detailed tags to APIGateway config.

manolama:
For some specific accounts that need the detail.
Also add cache and backend latency metrics.


brharrington:
This should also get ported to main branch to avoid confusion.

manolama:
Yup, just wanted to make sure it's working before porting.

PR #503, opened 2023-11-03
add clusters path to filter slotting data

copperlight:
The `/api/v1/clusters/:clusterName` path provides both an index view and a
verbose view of the stored slotting data. This view performs a Frigga cluster
name match on the AutoScalingGroup name and returns the result. This allows
a user to distinguish between `app-foo` and `app-foobar` when filtering.

The `/api/v2` paths which provided Edda compatibility have been removed. They
never implemented true compatibility, not having selectors, and thus were
misleading and made the API routes more complicated than necessary. There is
no usage of these endpoints outside of a few manual testing calls.

### DIFF
--- a/atlas-slotting/src/main/scala/com/netflix/atlas/slotting/SlottingApi.scala
+++ b/atlas-slotting/src/main/scala/com/netflix/atlas/slotting/SlottingApi.scala
@@ -15,6 +15,11 @@
  */
 package com.netflix.atlas.slotting
 
+import com.netflix.atlas.json.Json
+import com.netflix.atlas.pekko.CustomDirectives.*
+import com.netflix.atlas.pekko.WebApi
+import com.netflix.spectator.ipc.ServerGroup
+import com.typesafe.scalalogging.StrictLogging
 import org.apache.pekko.actor.ActorSystem
 import org.apache.pekko.http.caching.LfuCache
 import org.apache.pekko.http.caching.scaladsl.Cache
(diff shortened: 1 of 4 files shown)


atlas-slotting/src/main/scala/com/netflix/atlas/slotting/SlottingApi.scala
```diff
@@ -162,6 +154,30 @@ object SlottingApi {
...
 
+  def indexClusterList(
+    compress: Boolean,
+    slottingCache: SlottingCache,
+    clusterName: String
+  ): HttpResponse = {
+    mkResponse(
+      compress,
+      StatusCodes.OK,
+      slottingCache.asgs.keySet.filter(_.startsWith(clusterName))
```
brharrington, line 165: Would startsWith match the user expectation? I worry it could be confusing if you have a cluster with no stack or a stack that is a subset of another. For example:

```
app-foo
app-foobar
```

A user searching for ASGs of `app-foo` might be surprised for `app-foobar` ASGs to be included since the cluster name is different.

brharrington (approved):

atlas-slotting/src/main/scala/com/netflix/atlas/slotting/SlottingApi.scala
copperlight: I was thinking about how this might break, and I did not quite get here. This is a perfect point. I will add a test for this condition and fix up the filtering.

PR #516, opened 2023-12-20
atlas slotting: Update add and delete exception logging strings.

manolama:


atlas-slotting/src/main/scala/com/netflix/atlas/slotting/SlottingService.scala
```diff
@@ -378,7 +378,7 @@ class SlottingService(
         logger.error(s"failed to assign slots, not updating item $name: ${e.getMessage}")
         registry.counter(slotsErrorsId.withTag("asg", name)).increment()
       case e: Exception =>
-        logger.error(s"failed to update item $name: ${e.getMessage}")
+        logger.error(s"failed to add item $name: ${e.getMessage}")
```
brharrington, line 381: `put` align better with the operation name.

brharrington (approved):

PR #517, opened 2023-12-21
atlas slotting: log objects in CAS failures.

manolama:


atlas-slotting/src/main/scala/com/netflix/atlas/slotting/SlottingService.scala
```diff
@@ -348,6 +349,12 @@ class SlottingService(
         registry.counter(slotsChangedId.withTag("asg", name)).increment()
       }
     } catch {
+      case cas: ConditionalCheckFailedException =>
+        logger.error(
+          s"failed to update item $name: ${cas.getMessage}\n  " +
+            s"oldData: ${Util.decode(oldData)}\n  newData: ${Util.decode(newData)}"
```
brharrington, line 355: I'm not sure how the buffers might have been modified for the call. They may need to be reset before trying to read the contents.

I generally recommend against new lines in log messages. They make it much harder to grep the logs and can lead to other confusion.

manolama: ack, dropped the newlines. The buffers should be fine since the equality check will reset them.

brharrington: The equality check for ByteBuffer doesn't reset them. Looking at it closer, I think this logic is broken and it will not show the old data correctly, but more importantly may not be passed correctly to the request.

The `mkNewDataMergeSlots` calls `Utils.decompress(oldData)`. Internally that calls `Utils.toByteArray` which clears the buffer after copying to a byte array. So the `oldData` buffer will be empty after that call. So the `newData == oldData` check later works, but is maybe a bit confusing as it works because the `oldData` reference is returned if there isn't a change.

One concern, is if the else condition is hit, the `oldData` buffer is empty but is used for `updateAsgItemRequest` as one of the parameters. The second concern, is it will not show the original `oldData` buffer, though I suppose it will show what was used for the request.

atlas-slotting/src/main/scala/com/netflix/atlas/slotting/SlottingService.scala
```diff
@@ -348,6 +349,12 @@ class SlottingService(
         registry.counter(slotsChangedId.withTag("asg", name)).increment()
       }
     } catch {
+      case cas: ConditionalCheckFailedException =>
+        logger.error(
+          s"failed to update item $name: ${cas.getMessage}. " +
+            s"oldData: ${Util.decode(oldData)}, newData: ${Util.decode(newData)}"
+        )
```
brharrington, line 356: Can you include the exception in the log? That would provide a stack trace so it is clear where which branch failed.

manolama:
Superseded by **https://github.com/Netflix-Skunkworks/iep-apps/pull/518**

PR #518, opened 2023-12-22
slotting: avoid ByteBuffer conversions

brharrington:
Just keep the data as String/Array[Byte] and avoid unecessary conversions to and from ByteBuffer. All operations were first converting to intermediate byte array or other type anyway.

Also update log messages to properly log the exception so the stack trace will be present in the logs.

### DIFF
--- a/atlas-slotting/src/main/scala/com/netflix/atlas/slotting/DynamoOps.scala
+++ b/atlas-slotting/src/main/scala/com/netflix/atlas/slotting/DynamoOps.scala
@@ -15,7 +15,6 @@
  */
 package com.netflix.atlas.slotting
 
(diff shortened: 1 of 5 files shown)


copperlight:
This is a lot cleaner, especially with moving the `Gzip` compression to the boundary of interactions with DynamoDB, and exposing the full stack traces for DynamoDB operations.

I do not recall why we were originally using `ByteBuffer` internally, but working with `String` and `Array[Byte]` is certainly easier to reason about and less likely to produce the kind of gotcha behavior that was noted here.

brharrington:
I have only run some limited tests. Will test more and carefully roll out sometime in January.

PR #519, opened 2023-12-28
atlas slotting: more unit tests for Grouping.MkNewDataMergerSlots.

manolama:


atlas-slotting/src/main/scala/com/netflix/atlas/slotting/Grouping.scala
```diff
@@ -366,7 +366,7 @@ trait Grouping extends StrictLogging {
     oldData: String,
     newAsgDetails: AsgDetails,
     instanceInfo: Map[String, Ec2InstanceDetails],
-    slotsErrors: Counter
+    slotsErrors: () => Counter
```
brharrington, line 369: Why does this need to be a function? From the diff is seems like passing in the counter is sufficient.

manolama: Oh just a premature optimization to avoid instantiating a counter object when most of the calls are successful. Though creating the function isn't that much less expensive. I'll revert that bit.

brharrington (approved):

PR #520, opened 2023-12-29
slotting: only update slots from the crawl ASG task

brharrington:
Before both crawl instances and crawl ASGs would attempt to update slots if data was available. If both threads run the updateSlots operation at the same time, then it can cause issues including disabling some ASGs.

### DIFF
--- a/atlas-slotting/src/main/scala/com/netflix/atlas/slotting/SlottingService.scala
+++ b/atlas-slotting/src/main/scala/com/netflix/atlas/slotting/SlottingService.scala
@@ -168,8 +168,6 @@ class SlottingService(
       logger.info(s"crawled ${instanceInfo.size} instances in ${fmtTime(elapsed)}")
       lastUpdateInstances.set(clock.wallTime())
     }
-
-    if (allDataAvailable) updateSlots()
   }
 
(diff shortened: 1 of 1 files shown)


copperlight (approved):
I like this change and I think it will be a solid step towards making the final results more stable (if not solving for it entirely):

* The `crawlInstancesTask` will no longer attempt to update slots. It becomes the sole responsibility of the `crawlAsgsTask`, so multiple threads are not doing the same work. The ASG crawling task should be faster than the instance crawling, because there are fewer of them, and we should not see an increased delay in calculating slot changes.
* The `updateSlots` method captures the current state of `instanceInfo`, which helps make the streaming data processing changes to the DynamoDB table atomic for the iteration. The class state flags `asgsAvailable` and `instanceInfoAvailable` are reset to false early, which should help delay subsequent updates, and give more time for the current iteration to complete.
* The `updateItem` method accepts the snapshot of the `instanceInfo` class variable, for consistency.

I think the earlier logging changes should help us keep track of whether or not we are still seeing DynamoDB update inconsistencies after this change lands.

I am not sure if should have an error counter that increments when ASGs are dropped from the result set - this is a common thing that happens when ASGs are rotated, but this is also how we observed this flapping behavior.

PR #548, opened 2024-05-01
Add ExecutedVersion Dimension and three new PC metrics

mpalriwal-Netflix:


atlas-cloudwatch/src/main/resources/lambda.conf
```diff
@@ -26,10 +26,36 @@ atlas {
 
       dimensions = [
         "FunctionName",
-        "Resource"
+        "Resource",
+        "ExecutedVersion"
```
manolama, line 30: Add this tag to the `reference.conf` mapper as something like `aws.executed.version`

manolama (changes requested):

atlas-cloudwatch/src/main/resources/lambda.conf
mpalriwal-Netflix: Updated

manolama (approved):

PR #552, opened 2024-05-03
Add aws.metric tag to filter metric counter

mpalriwal-Netflix:
Add aws.metric tag to filter metric counter, to confirm which all metrics names are getting filtered out

### DIFF
--- a/atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchMetricsProcessor.scala
+++ b/atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchMetricsProcessor.scala
@@ -176,12 +176,24 @@ abstract class CloudWatchMetricsProcessor(
         case Some(namespaceRules) =>
           namespaceRules.get(dp.metricName) match {
             case None =>
-              registry.counter(filteredMetric.withTag("aws.namespace", dp.namespace).withTag("aws.metric", dp.metricName)).increment()
+              registry
+                .counter(
+                  filteredMetric
+                    .withTag("aws.namespace", dp.namespace)
+                    .withTag("aws.metric", dp.metricName)
+                )
+                .increment()
               debugger.debugIncoming(dp, IncomingMatch.DroppedMetric, receivedTimestamp)
             case Some(tuple) =>
               val (category, _) = tuple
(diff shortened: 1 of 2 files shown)


atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchMetricsProcessor.scala
```diff
@@ -176,12 +176,24 @@ abstract class CloudWatchMetricsProcessor(
         case Some(namespaceRules) =>
           namespaceRules.get(dp.metricName) match {
             case None =>
-              registry.counter(filteredMetric.withTag("aws.namespace", dp.namespace).withTag("aws.metric", dp.metricName)).increment()
+              registry
+                .counter(
+                  filteredMetric
+                    .withTag("aws.namespace", dp.namespace)
+                    .withTag("aws.metric", dp.metricName)
```
brharrington, line 183: How often would this actually be useful? It seems like a debug or maybe trace level log would probably be sufficient and it could ben enabled if actually debugging something. 

As long as the set of names is strictly bounded to a small set it should be ok. I guess that should be true for CW.

brharrington (approved):
Test case needs to be fixed.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchMetricsProcessor.scala
mpalriwal-Netflix: Agree even a debug level log will work, while the same tag is already used for ingestion/published, so this will bring parity across all operation  

[filtered](https://go.prod.netflix.net/qxhqV8)

mpalriwal-Netflix:
> Test case needs to be fixed.

@brharrington  fixed

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchMetricsProcessor.scala
manolama: It's been particularly useful for debugging latency in publishing and receipt. The cardinality is fairly low and stable so I'm not too worried.

PR #560, opened 2024-07-18
try using new sonatype publishing host

brharrington:
See if it clears up publishing issues.


brharrington:
Doesn't support snapshots: https://central.sonatype.org/faq/snapshot-releases/

PR #574, opened 2024-08-29
Add additional context to druidQueries to improve debuggability and linking Atlas Queries to specific Druid Queries

ajamato:


ajamato:
@bsyk Would you mind doing an initial review please?

atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
```diff
@@ -567,9 +569,17 @@ object DruidDatabaseActor {
     }
   }
 
+  def toDruidQueryContext(request: DataRequest): Map[String, String] = {
+    Map(
+      "atlasQuerySource"     -> request.config.map(_.id).getOrElse("unknown"),
+      "atlasQueriesCombined" -> request.exprs.mkString(" "),
```
bsyk, line 575: Is it possible to get the original query rather than trying to combine the resulting data-expr from each query?
By original I mean the query sent to Atlas that was interpreted and split into (potentially multiple) expression(s) that are then translated into Druid queries.

atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
```diff
@@ -567,9 +569,17 @@ object DruidDatabaseActor {
     }
   }
 
+  def toDruidQueryContext(request: DataRequest): Map[String, String] = {
```
bsyk, line 572: We'd also talked about trying to get hold of the requester's IP address so we could have some tracing at the Druid layer. Is that possible to grab and add to the context here?

atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
brharrington: `request.config.fold("unknown")(_.query)`

ajamato: Oh, I didn't realize that it gets split prior to this. Likely I can poke around the debugger and figure out a way to do that... Though it may mean going much higher up in the stack to plumb the information through.

atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
ajamato: Ip does not seem like it will be an option / available. I propose we start with this and see if we can enchance the `id` field to be more specific, or introduce a new field i.e. `humanReadableQueryName` in addition to `id`

bsyk (approved):

atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
ajamato: Thanks. I did this to be congruent with the existing code style, which I believe is logically equivalent to using fold.

`      "atlasQueryString" -> request.config.map(_.query).getOrElse("unknown"),`

brharrington (approved):

PR #577, opened 2024-10-02
Update poller to consider HRM fast-polling for Media Live/Connect NS

mpalriwal-Netflix:
Update poller to consider high resolution metric fast-polling for (Media Live/Connect) specific namespace


manolama (approved):
Just need to see if the 5s polling pans out. There may be improvements needed there.

PR #581, opened 2024-10-07
Use AtlasRegistry for High Resolution CW Polled Metrics

mpalriwal-Netflix:
* Defined eval/config uri for AtlasRegistry to publish to LWC, specific to owner account:stack:region. 
* Poller use `hrmFrequency` to setup a fast polling scheduler same as ``period``. 
* Poller publish collected metric dp to ``sendToRegistry`` method only for period < 60
* ``sendToRegistry`` will call PublishRouter ``publishToRegistry`` which transform FirehoseMetric to a `gaugeValue` and call its own queue from accountMap, based on account/region value. 
* PublishQueue is one per account/region/stack (and one default main) which will now allow ``updateRegistry`` method to invoke ``PublishClient`` registry specific updates.


atlas-cloudwatch/src/main/resources/reference.conf
```diff
@@ -116,7 +116,6 @@ atlas {
       // execute.
       frequency = 1m
 
-      // frequency for high resolution metrics
```
brharrington, line 119: Why remove the comment?

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchMetricsProcessor.scala
```diff
@@ -457,6 +458,29 @@ abstract class CloudWatchMetricsProcessor(
...
+    datapoint: FirehoseMetric,
+    category: MetricCategory,
+    receivedTimestamp: Long
+  ): Unit = {
+
+    category.metrics.foreach(md => {
+      val metric = MetricData(
+        MetricMetadata(category, md, toAWSDimensions(datapoint)),
+        None,
+        Option.apply(datapoint.datapoint),
```
brharrington, line 471: `Option(datapoint.datapoint)`

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchMetricsProcessor.scala
```diff
@@ -32,6 +32,7 @@ import software.amazon.awssdk.services.cloudwatch.model.Metric
 
 import java.time.Instant
 import java.util
+import java.util.Optional
```
brharrington, line 35: This appears to be unused

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/PublishRouter.scala
```diff
@@ -124,12 +150,32 @@ class PublishRouter(
...
-  private[cloudwatch] def getQueue(datapoint: AtlasDatapoint): Option[PublishQueue] = {
-    datapoint.tags.get("nf.account") match {
+  /**
+   * Routes the data to the proper atlas registry instance based on the `nf.account` and region tag.
+   *
+   * @param datapoint
+   *     The non-null data point.
+   */
+  def publishToRegistry(datapoint: AtlasDatapoint): Unit = {
+    val atlasGaugeDp = toGaugeValue(datapoint)
```
brharrington, line 160: Why is this conversion needed?

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/poller/GaugeValue.scala
```diff
@@ -0,0 +1,37 @@
...
+import com.netflix.spectator.api.Statistic
+
+case class GaugeValue(tags: Map[String, String], value: Double) {
+
+  def id: Id = {
+    import scala.jdk.CollectionConverters._
+    val ts = (tags - "name").map {
+      case (k, v) => k -> (if (v.length > 120) "VALUE_TOO_LONG" else v)
+    }.asJava
+    Id.create(tags("name")).withTags(ts).withTag(Statistic.gauge)
```
brharrington, line 28: The statistic is typically added/managed by the registry, not directly by the user.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/poller/PublishClient.scala
```diff
@@ -0,0 +1,116 @@
...
+
+  private val maxMeters = super.maxNumberOfMeters()
+
+  override def get(k: String): String = {
+    val prop = s"atlas.cloudwatch.account.routing.$k"
+    if (config.hasPath(prop)) config.getString(prop) else null
+  }
+
+  override def uri: String = {
+    if (uri == null) "http://localhost:7101/api/v1/publish"
```
brharrington, line 77: The result of this condition is ignored

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/PublishQueue.scala
```diff
@@ -90,14 +101,13 @@ class PublishQueue(
...
-    *
-    * @param datapoint
-    *     The non-null data point to enqueue.
-    */
+
+  logger.info(
+    s"Setup queue for stack ${stack} publishing URI ${uri}, lwc-config URI ${configUri}, eval URI ${evalUri}"
+  )
+
+  def updateRegistry(g: GaugeValue): Unit = registryPublishClient.updateGauge(g.id, g.value)
```
brharrington, line 109: Are all the types really gauges? I would have expected at least some counters.

build.sbt
```diff
@@ -1,5 +1,5 @@
-
-lazy val `iep-apps` = project.in(file("."))
+lazy val `iep-apps` = project
```
brharrington, line 1: All the reformatting makes it hard to see what the actual change is in this file. Please avoid reformatting as part of the change.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/PublishQueue.scala
```diff
@@ -90,14 +103,35 @@ class PublishQueue(
...
+  private def toDoubleValue(
+    datapoint: AtlasDatapoint
+  ): DoubleValue = {
+    DoubleValue(datapoint.tags, datapoint.value)
+  }
+
+  private def toCounterValue(
+    datapoint: AtlasDatapoint
+  ): LongValue = {
+    LongValue(datapoint.tags, datapoint.value.longValue)
```
brharrington, line 132: Why are you converting to a long? That could truncate the value.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/poller/PublishClient.scala
```diff
@@ -0,0 +1,110 @@
...
+class PublishClient(config: PublishConfig) extends AbstractService {
+
+  private val publishRegistry = new AtlasRegistry(Clock.SYSTEM, config)
+
+  def updateGauge(id: Id, value: Double): Unit = {
+    publishRegistry.maxGauge(id).set(value)
+  }
+
+  def updateCounter(id: Id, value: Long): Unit = {
+    publishRegistry.counter(id).increment(value)
```
brharrington, line 39: I would suggest using `add(value)` and keeping it as a double. Also is the value passed in here the number of of updates? If it has already been converted to a rate you may need to undo before sending to the counter. Otherwise the result will be a rate of rates.

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/PublishQueue.scala
mpalriwal-Netflix: switching to ``add(value) ``

atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchMetricsProcessor.scala
```diff
@@ -862,8 +862,17 @@ abstract class CloudWatchMetricsProcessor(
       // TODO - clean this out once tests are finished
       // (if (testMode) s"TEST.${definition.alias}" else definition.alias))
       definition.alias)
-    val newValue =
-      definition.conversion(metric.meta, metric.datapoint(Instant.ofEpochMilli(timestamp)))
+
+    // Determine if rate conversion should be applied based on the step value
+    val applyRateConversion = step >= 60
```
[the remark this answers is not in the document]
mpalriwal-Netflix: I originally kept it as part of MetricDefn, so if needed we can apply it per metric level too, but agree for now moving to ``rate`` method locally will be more simpler, as we need for all MetricDefn.

PR #591, opened 2024-10-17
Simplify polled cw-dp to atlas-dp match/conversion

mpalriwal-Netflix:
* ``sendToRegistry `` does not require full category scan to match a given metric-name, we can pass the existing metadata from poller, to directly transform to an Atlas datapoint.

* Added some more metric/tag/logs for debugging purpose only, will remove later as this will be 5s frequency noise.

### DIFF
--- a/atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchPoller.scala
+++ b/atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchPoller.scala
@@ -118,8 +118,13 @@ class CloudWatchPoller(
         val categories = map.getOrElse(offset, List.empty)
         map += offset -> (categories :+ category)
         logger.info(
-          s"Setting offset of ${offset}s for ns ${category.namespace} period ${category.period}"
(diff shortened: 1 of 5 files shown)


atlas-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchPoller.scala
```diff
@@ -118,8 +118,13 @@ class CloudWatchPoller(
...
         map += offset -> (categories :+ category)
         logger.info(
-          s"Setting offset of ${offset}s for ns ${category.namespace} period ${category.period}"
+          s"Setting offset of ${offset}s for ns ${category.namespace} period ${category.period} for ${category.metrics.size} definition"
         )
+        category.metrics.foreach(md => {
+          logger.info(
+            s"${category.namespace} :: step: ${category.period}s, name: ${md.name}, tags: ${md.tags.keySet}"
+          )
+        })
```
mpalriwal-Netflix, line 127: Just for debugging purpose to confirm "timer" category are getting three defn or not

PR #595, opened 2024-10-17
Add new NLB/ALB metrics

mpalriwal-Netflix:


atlas-cloudwatch/src/main/resources/alb.conf
```diff
@@ -40,6 +40,16 @@ atlas {
...
           conversion = "sum,rate"
         },
+        {
+          name = "IPv6ProcessedBytes"
+          alias = "aws.alb.ipv6ProcessedBytes"
+          conversion = "sum,rate"
+        },
+        {
+          name = "IPv6RequestCount"
+          alias = "aws.alb.ipv6ProcessedCount"
```
brharrington, line 50: This should should be named to be more consistent with the existing mapping for `RequestCount`. Also, this metric looks to be in the wrong section, it should probably go in `alb-tg-zone`.

Ideally it would be modeled such that the IP address type would be a dimension that you would drill into rather than a different name. Unfortunately it looks like AWS doesn't report IPv4 individually and lumps both types in `RequestCount` and `ProcessedBytes`.

mpalriwal-Netflix: Thanks @brharrington , this metric does not come with ``TargetGroup`` dimension, so I skip ``alb-tg-zone`` category.  For naming, you are right I'll switch it to ``aws.alb.ipv6Requests``

PR #600, opened 2024-10-31
Allow an account specific allow list for some categories

mpalriwal-Netflix:
This change will allow to restrict some high cardinality or specific custom namespace metric for only a defined allowed list of account, unlike all other namespace is allowed for all as default. 

ec2-Asg config will only allow ASG dimension, which is low cardinality and should be fine for all accounts.


atlas-cloudwatch/src/main/resources/reference.conf
```diff
@@ -542,6 +542,7 @@ atlas {
       "dynamodb-table-ops",
       "dynamodb-capacity",
       "dynamodb-replication",
+      "ec2-asg",
```
brharrington, line 545: Can you keep this list sorted? It makes it easier to sanity check later.

PR #609, opened 2024-12-10
Create MSK Serverless scrape config

brituck:
InfraAPI is using AWS MSK Serverless for event streaming. Adds MSK serverless metrics config for atlas. See https://docs.aws.amazon.com/msk/latest/developerguide/serverless-monitoring.html for metrics details.

### DIFF
--- a/atlas-cloudwatch/src/main/resources/msk-serverless.conf
+++ b/atlas-cloudwatch/src/main/resources/msk-serverless.conf
@@ -1,0 +1,104 @@
+atlas {
+  cloudwatch {
+
+    // https://docs.aws.amazon.com/msk/latest/developerguide/serverless-monitoring.html
+    msk-serverless-cluster = {
+      namespace = "AWS/Kafka"
+      period = 1m
+      end-period-offset = 5
+
+      dimensions = [
+        "Cluster Name",
+        "Topic"
+      ]
+
+      metrics = [
+        {
+          name = "BytesIn"
+          alias = "aws.mskserverless.bytes"
+          conversion = "sum,rate"
+          tags = [
+            {
+              key = "id"
+              value = "in"
+            }
+          ]
+        },
+        {
+          name = "BytesOut"
+          alias = "aws.mskserverless.bytes"
+          conversion = "sum,rate"
+          tags = [
+            {
+              key = "id"
+              value = "out"
+            }
+          ]
+        },
+        {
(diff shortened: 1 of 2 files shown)


atlas-cloudwatch/src/main/resources/msk-serverless.conf
```diff
@@ -0,0 +1,104 @@
...
+  cloudwatch {
+
+    // https://docs.aws.amazon.com/msk/latest/developerguide/serverless-monitoring.html
+    msk-serverless-cluster = {
+      namespace = "AWS/Kafka"
+      period = 1m
+      end-period-offset = 5
+
+      dimensions = [
+        "Cluster Name",
```
brituck, line 11: Is this correct with the space? That's how it's listed in AWS, but not sure if we do some sort of whitespace removal

atlas-cloudwatch/src/main/resources/msk-serverless.conf
```diff
@@ -0,0 +1,104 @@
+atlas {
+  cloudwatch {
+
+    // https://docs.aws.amazon.com/msk/latest/developerguide/serverless-monitoring.html
+    msk-serverless-cluster = {
+      namespace = "AWS/Kafka"
+      period = 1m
+      end-period-offset = 5
```
brituck, line 8: MSK publishes these metrics once per minute. Do we need an offset? Was following other per-minute metrics configs like Dynamo

manolama: I don't think we use it now but leave it in and we'll clean it up later.

atlas-cloudwatch/src/main/resources/msk-serverless.conf
```diff
@@ -0,0 +1,104 @@
...
+
+    msk-serverless-consumergroup = {
+      namespace = "AWS/Kafka"
+      period = 1m
+      end-period-offset = 5
+
+      dimensions = [
+        "Cluster Name",
+        "Topic",
+        "Consumer Group"
```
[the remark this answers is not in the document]
manolama: Also, add these three to the tag map in https://github.com/Netflix-Skunkworks/iep-apps/blob/main/atlas-cloudwatch/src/main/resources/reference.conf#L236 with dotted notation for Atlas please.

manolama (approved):

PR #615, opened 2025-01-13
Dual Publish to Valkey and Redis via Jedis client only

mpalriwal-Netflix:
``updateCache`` method will do dual write 

Currently we are using Jedis only, but once we completely move to Valkey, we will switch to valkey-java client https://github.com/valkey-io/valkey-java


manolama (approved):
I think it's ok.

PR #642, opened 2025-05-12
Update DruidDatabaseActor to query using the max of all metrics' step sizes and the request step size

ajamato:


atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
```diff
@@ -88,9 +82,11 @@ class DruidDatabaseActor(config: Config, service: DruidMetadataService)
     context.system.scheduler.scheduleAtFixedRate(0.seconds, 10.minutes, self, Tick)
 
   def receive: Receive = {
-    case Tick        => refreshMetadata(sender())
-    case m: Metadata => metadata = Metadata(m.datasources.filter(_.nonEmpty))
-
+    case Tick => refreshMetadata(sender())
+    case m: Metadata => {
+      metadata = Metadata(m.datasources.filter(_.nonEmpty))
+      sender() ! "metadata_updated"
```
ajamato, line 88: I added the response here, so I could wait for it in my test before issuing a query.
Let me know if this approach is okay or not. It is unclear to me if this will affect senders who may receive this response.

atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
```diff
@@ -335,7 +331,9 @@ class DruidDatabaseActor(config: Config, service: DruidMetadataService)
 
     val druidQueries = toDruidQueries(metadata, druidQueryContext, fetchContext, expr).map {
       case (tags, metric, groupByQuery) =>
-        val druidStep = math.max(metric.primaryStep, fetchContext.step)
+        // Always use the step set on the druid metric from the datasource metadata.
+        // Ignore the "step" parameter on the request.
```
bsyk, line 335: I don't think we want to ignore the requested step. We should still honor the step if it's larger/more corse than what the datasource can supply.

brharrington: Correct, it should still be taking the max and honoring the context step.

build.sbt
```diff
@@ -78,6 +78,10 @@ lazy val `atlas-druid` = project
     Dependencies.log4jApi,
     Dependencies.log4jCore,
     Dependencies.log4jSlf4j,
+
+    Dependencies.pekkoTestkit % "test",
+    Dependencies.mockitoCore % "test",
+    Dependencies.mockitoScala % "test",
```
brharrington, line 84: We're trying to phase out mockito usage in this repo and don't want to introduce more. At first glance it doesn't appear to really be needed for these tests.

atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
```diff
@@ -335,7 +331,9 @@ class DruidDatabaseActor(config: Config, service: DruidMetadataService)
...
+        val druidStep = metric.primaryStep
```
brharrington, line 336: Looking at this a bit further, it would need to be done at a higher level as we want agreement across all signals. Doing it here could have a mix if multiple signals with different steps sizes are used. Also, increasing the step size within the actor isn't going to work right now because it isn't known in the response leading to problems with miss-allignment of the values to the time axis.

build.sbt
ajamato: I do need to mock out the responses see `newDruidClient` in `DruidDatabaseActorSuite` and `DruidClientSuite`

If there is a different mock tool you recommend I can migrate to that it. But this was what I saw being used currently in the project.

atlas-druid/src/main/scala/com/netflix/atlas/druid/DruidDatabaseActor.scala
brharrington: There was a change needed in the actor, but also the consumer of the response. In the actor the step needs to be determined before fanning out for each expression so it will work correctly across independent expressions as well as multiple primary steps for metrics matching a single expression. I have an alternate patch, #648, that seems to be working.

ajamato:
Closing, in favor of

https://github.com/Netflix-Skunkworks/iep-apps/pull/648/files#diff-ddbafd573b97a3729eaf7f4c3d533e7b98373018b417940ccc2a9a6ea21f70f9

PR #735, opened 2026-06-10
avoid RefIntHashMap.foreach

mpalriwal-Netflix:
Replace the foreach call with toMap.keys.toArray


atlas-aggregator/src/main/scala/com/netflix/atlas/aggregator/ShardedAggregatorService.scala
```diff
@@ -180,12 +180,7 @@ object ShardedAggregatorService {
         stringTable.put(v, 0)
       }
     }
-    val strings = new Array[String](stringTable.size)
-    var i = 0
-    stringTable.foreach { (k, _) =>
-      strings(i) = k
-      i += 1
-    }
+    val strings = stringTable.toMap.keys.toArray
```
brharrington, line 183: This would increase allocations a lot as it creates an intermediate map.

